PR #1066, opened 2026-06-14
fix(cli): honor glm.models from llm.yaml in moai glm (Fixes #1065)

### LINKED ISSUE #1065: moai glm does not reflect glm.models from .moai/config/sections/llm.yaml after GLM 5.2 update
### Description
After the GLM 5.2 update, changing model names in `.moai/config/sections/llm.yaml` does not take effect. The `moai glm` command continues to use hardcoded default models (glm-5.1 for opus, glm-4.7 for sonnet, glm-4.5-air for haiku) instead of reading from the yaml configuration.

### Root Cause Analysis
The `loadGLMConfig` function in `internal/cli/glm.go:628` has a fallback logic that always uses `config.NewDefaultLLMConfig()` when `deps.Config` is nil or empty. This means the user's llm.yaml settings are ignored, and the hardcoded defaults from `internal/config/defaults.go:43-45` are always applied.

### Expected Behavior
- `moai glm` should read `glm.models` from `.moai/config/sections/llm.yaml`
- Users should be able to override default models (e.g., use glm-5.2[1m] instead of glm-5.1) via configuration file

### Actual Behavior
- Model changes in llm.yaml are completely ignored
- Probing test confirmed: changing glm.models.high to "glm-PROBE-HIGH" still injects "glm-5.1"

### Workaround
- Use `.zshrc` function `glm-cc()` that directly exports ANTHROPIC_DEFAULT_*_MODEL before exec claude
- Or rebuild moai binary after modifying `internal/config/defaults.go`

### Priority
High (blocks model configuration workflow)

### Environment
- moai-adk version: 2.14.0 (installed binary) vs 2.8.2 (local source)
- Platform: macOS (Darwin 25.5.0)
- Shell: zsh

GoosLab:
## Summary

Fixes #1065 — `moai glm` ignored `glm.models` from `.moai/config/sections/llm.yaml` and always injected hardcoded default models.

## Root cause (confirmed vs. the reporter's claim)

The reporter's cited line (`glm.go:628`) and their `glm-5.1` symptom are from a **stale installed binary** — current source already defaults `high` to `glm-5.2[1m]`. But the underlying "yaml models ignored" defect is **real** and was traced to a deeper cause than the reporter's hypothesis:

1. **Primary defect (deeper than reported):** the `moai glm` runtime path `runGLM → unifiedLaunch → applyGLMMode → loadGLMConfig` **never calls `deps.Config.Load()`**. So `deps.Config.Get()` returns `nil` at runtime and `loadGLMConfig` always fell through to `config.NewDefaultLLMConfig()` — discarding the user's `llm.yaml` entirely, regardless of any field.
2. **Secondary defect:** the in-memory branch was additionally gated on `cfg.LLM.GLM.BaseURL != ""`, so even a *loaded* config with custom `models` but no explicit `base_url` was discarded.

## Fix

`loadGLMConfig` now reads `llm.yaml` directly from the project root via the existing `loadLLMSectionOnly` helper when the in-memory config is unloaded (the live runtime case), and model resolution is decoupled from the `base_url` gate (default `base_url` fills in when omitted). `resolveGLMModels` and the default model values in `internal/config/defaults.go` are **unchanged**.

## Test Plan

Reproduction-first (RED → GREEN). New tests in `internal/cli/glm_new_test.go`:

- `TestLoadGLMConfig_ReadsDiskModelsWhenConfigUnloaded` — issue #1065 scenario: custom models, no `base_url`, unloaded `ConfigManager`. **RED** before fix (injected `glm-5.2[1m]` instead of `glm-PROBE-HIGH`), **GREEN** after.
- `TestLoadGLMConfig_ReadsDiskModelsWithNilDeps` — disk models honored with `deps == nil`.
- `TestLoadGLMConfig_DiskBaseURLHonored` — explicit disk `base_url` preserved.
- `TestLoadGLMConfig_AbsentLLMYAMLFallsToDefaults` — no `llm.yaml` → system defaults.

Verification: `go test ./...` (all pass), `go vet`, `golangci-lint run` (0 issues), `GOOS=windows GOARCH=amd64 go build ./...`, `go test -race ./internal/cli/...`. C-HRA-008 subagent-boundary grep clean (no `AskUserQuestion`).

## Note for the reporter

The `glm-5.1` you observed is from a **stale 2.14.0 binary** — rebuild/reinstall (`make install` or latest release) to get the `glm-5.2[1m]` defaults. This PR lands the real defect: your `glm.models` in `llm.yaml` are now honored at runtime.

Fixes #1065

🗿 MoAI <email@mo.ai.kr>


GoosLab:
## ✅ Review: Approve (rebase before merge)

Solid, well-scoped fix for #1065. The `runGLM → unifiedLaunch → applyGLMMode` path never calls `deps.Config.Load()`, so `deps.Config.Get()` returns nil and user `llm.yaml` models were silently dropped in favor of hardcoded defaults.

**What works well:**
- `switch` cleanly prefers populated in-memory config, falls back to a direct disk read (`loadLLMSectionOnly`) on the live `glm` runtime path.
- `base_url` / `env_var` empty-value default fill-in is correct and independent of model resolution.
- Three tests cover the matrix precisely: config-unloaded, nil-deps, and explicit-disk-base_url — including an inline RED-reproduction note.

**Note:** The failing `Test (*)` checks are **not** caused by this PR — they are a stale-base RED (`TestClose_FullClose_ProducesCommit` in `internal/spec`, already green on latest `main`). A rebase onto `main` will clear them.

PR #1128, opened 2026-07-23
fix(windows): release-verify stability — atomicfile, advisory lock, checkpoint

GoosLab:
## Summary
Windows test-portability and release-verify stability fixes (canonical branch; supersedes the fix/windows-test-portability, -round2, and -shortname-slug variants which are superseded/duplicate).

- route session-store checkpoint writes through atomicfile
- make advisory lock survive delete-pending; bound the retry narrowly
- guard terminal background query behind a non-TTY check
- route spec-close lock through delete-pending transient retry

🗿 MoAI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,10 @@ jobs:
       - name: Run tests with race detector (Windows, no coverage)
         if: matrix.os == 'windows-latest'
         shell: bash
-        run: go test -race ./...
+        # -timeout 20m raises Go's 10m per-package default; internal/cli under
+        # -race on Windows sits at that boundary. Kept in sync with the same
+        # step in release-pr-multi-os.yml.
+        run: go test -race -timeout 20m ./...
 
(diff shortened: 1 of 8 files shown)


GoosLab:
Superseded by main — closing without merge.

Verification (branch tip vs current origin/main, direct 2-way file diff):
- internal/migration/version.go — identical to main (0 diff)
- internal/migration/version_windows.go — identical to main (0 diff)
- internal/session/store.go — identical to main (0 diff)
- internal/spec/lock_windows.go — identical to main (0 diff)
- internal/tui/detect.go / detect_test.go — branch is an OLDER version that removes the `IsDark` function main already has; merging would regress.

The lock/atomicfile/session/migration work on this branch already landed on main through other PRs (byte-identical). The only remaining delta is a stale detect.go that predates main's current state. Net value is zero and it would regress `IsDark`, so this is closed rather than merged. The redundant siblings (fix/windows-test-portability, -round2, -shortname-slug) are likewise superseded.

PR #1294, opened 2026-08-02
perf(hook): parallelize SessionStart + defer advisory scans to cut input lag

GoosLab:
## What
SessionStart hook ran ~12 steps serially; Claude Code waited for completion before accepting keyboard input — the "TUI up, keys blocked, then unblock" symptom. This is the Rank-1 fix from a launcher input-lag bottleneck survey.

## Changes (internal/hook/ only)
- **errgroup** the independent synchronous steps (settings chain, registry, skill symlinks, migration) into 4 concurrent tasks
- **defer** heavy advisory scans (telemetry prune, stale-memory, proposals, status drift) to a background goroutine joined with a short bounded deadline (250ms): common-case advisories land, slow scans drop idempotently
- **async/sync test seam** (TestMain sets sync) so the ~50 Handle-calling tests don't leak the goroutine; production keeps the async path
- **de-parallelize** the sole parallel os.Stderr writer test (pre-existing flake surfaced by the new goroutine)
- **maps.Copy** cleanup in mergeData

## Behavior preservation
Synchronous side effects unchanged: credentials, teammate-mode, tmux-env, multi-session registry write, goal-orphan move, skill symlinks, migration, AdditionalContext/session-id attribution. Deferred advisory Data keys are `json:"-"` (internal-only).

## Verification (independently reproduced, not just agent self-report)
- `go vet ./internal/hook/...` → 0
- `golangci-lint run ./internal/hook/...` → 0 issues
- `go test -race -count=10 ./internal/hook/` → ok, **0 races** (108.9s)
- `go test -race -count=3 ./internal/hook/` ×2 → ok, 0 races
- characterization tests (async path) green
- `go build ./...` → ok

## Notes
- Residual: per-step wall-clock not profiled; 250ms bound reasoned from drift 2s ceiling. Production goroutine lifetime inferred (idempotent), not measured.
- De-parallelizing `TestPostTool_MemoryMissingType` is the minimal deterministic fix; deeper `io.Writer` injection is out of scope.
- Branch is 2 commits behind main (rebase was permission-denied in-session); use **Update branch** if CI requires up-to-date before auto-merge.

🗿 MoAI

## Summary by CodeRabbit

* **Performance**
  * Session startup now completes more quickly by running independent setup tasks in parallel.
  * Advisory scans run in the background, with a bounded wait for results so startup is not unnecessarily delayed.

* **Reliability**
  * Startup continues to preserve settings updates, registry changes, migrations, and other returned data.
  * Added coverage for deferred scans, bounded completion, and synchronous side effects.


GoosLab (commented):
Independent review — 0 Critical, security-clean (no injection/authn/sensitive-data surfaces). Concurrent-map hazard correctly handled (task-local maps + `mergeData` converging after `g.Wait()`); settings-chain dependency (`ensureTmuxGLMEnv` reads `ensureGLMCredentials` token) preserved inside one task; registry RMW + `pruneGoalOrphans` sequenced inside Task 2; `-race` clean; all CI green (3-OS Test/Lint/CodeQL/claude-review/CodeRabbit).

Follow-ups worth tracking (non-blocking):
- codecov patch 63.36% concentrated on the deferred-scan branch (production async path)
- deferred-goroutine lifetime is inferred-idempotent, not measured
- `mergeData` keys have no documented cross-task ownership invariant

PR #1358, opened 2026-08-04
ci: spec-status-sync graceful skip on protected branch

GoosLab:
Under PR-mandatory (`enforce_admins:true`), the spec-status-auto-sync bot's `git push origin main` is rejected by branch protection → job fails (non-blocking but noisy; surfaced by #1341). This catches the protection-rejection sentinels and treats them as a graceful no-op (exit 0, SPEC status stays as-authored). Non-protection push errors still fail the job. Minimal change to the push step of `spec-status-auto-sync.yml`.

### DIFF
--- a/.github/workflows/spec-status-auto-sync.yml
+++ b/.github/workflows/spec-status-auto-sync.yml
@@ -120,5 +120,18 @@ jobs:
 
           # Push changes if any commits were made
           if [ "$(git rev-parse --short HEAD)" != "$(git rev-parse --short origin/main)" ]; then
-            git push origin main
+            # Under PR-mandatory (enforce_admins:true per CLAUDE.local.md §23), a direct
+            # push to main is rejected by branch protection. Treat that rejection as a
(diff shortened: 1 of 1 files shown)


GoosLab:
Closing as superseded: the workflow this PR patches — `.github/workflows/spec-status-auto-sync.yml` — was removed from `main` in #1369 (`11ee43412`), so the graceful-skip condition no longer has a target.

Verified:
```
$ git cat-file -e origin/main:.github/workflows/spec-status-auto-sync.yml
fatal: path ... does not exist in 'origin/main'
$ git merge-tree --write-tree origin/main f1858b9ede
CONFLICT (modify/delete): .github/workflows/spec-status-auto-sync.yml deleted in origin/main and modified in f1858b9
```

A rebase would only surface a modify/delete conflict with nothing to restore. If the status-sync workflow is ever reintroduced, the protected-branch skip should be designed into it from the start rather than revived from this branch.

PR #1523, opened 2026-08-14
refactor(web): retire the transitional layers — settings into the shell, board into /specs

GoosLab:
## What this closes

The console redesign (7e4f0eb66) landed the new shell for four screens and left two scaffolds behind. Both are gone now.

| Scaffold | Why it existed | What replaced it |
|---|---|---|
| `console.css` §12 compatibility layer | Kept the settings screen readable on the new tokens while its markup still used the old class names | The settings markup now speaks the §4-§7 vocabulary, so the layer and the §1 token aliases are deleted (`console.css` 429 → 387 lines, ends at §11) |
| `/specs/board` route | Held the close-debt list and MUST-FIX commands; nothing linked to it | Two panels on `/specs`, above the table |

Both had to go together: §12's `.page` / `.pagehead` / `.appbar` / `.banner--error` / `.board-*` rules were shared by the settings screen **and** the board page, so removing only one would have left half the layer alive.

## Settings moved into `Shell(vm)`

Its own `<html>`, appbar, hero band, horizontal tab strip, profile bar and bottom save bar are replaced by the rail, the top bar's context chips and save cluster, and the rail's vertical subnav.

- **Every panel stays in the DOM.** Rendering only the active tab would make the others' fields unsubmitted, which the parser reads as "preserve" — the atomic-save contract would break silently.
- **Tab switching stays client-side.** The subnav links to `/settings?tab=<id>` and the server honours it, so it works without JS; `app.js` upgrades the links to class toggling. That is not for speed — a full reload would discard unsaved edits, since all panels share one form. A hidden `__tab` field returns you to the same tab after saving.
- **One profile surface.** The settings-page profile bar folded into the rail popover, so switch / create / rename / delete are now reachable from every screen. Routes and field names (`profile_name`, `new_name`) unchanged.

## Contract preservation

Only class names and element placement moved. Unchanged: every `name=` attribute, form `action`, POST route, hidden `__present` companion, `data-i18n` key, option `value`, and `id`.

## Defects found along the way

1. **The appbar context line was hardcoded.** `lang: ko · model: claude-opus-5 · effort: high · dev: ddd` was a literal, so it read the same whatever the profile held. Its test passed only because the mock values happened to match the literal. The chips now come from the view model, and the test uses values that differ from it so a regression to a literal cannot pass.
2. **Synthesized tab i18n keys did not exist.** The shell built `"tab."+id`, but the dictionary has no `tab.identity` etc. — tab labels would have rendered blank in all four locales. Each tab now carries its own `LabelKey` (the existing `sec.<x>.title`).
3. **The profile popover had no JS wiring and could never open.** With the profile bar deleted, that would have made all four profile operations unreachable. `wirePopovers()` closes it; outside-click and Esc are registered once at IIFE top level rather than per htmx swap.

## Deliberate capability loss

The board's status-distribution summary (8 enum values + out-of-enum + unknown) is **not** reproduced. The `/specs` filter chips cover the four statuses that matter for filtering. Say the word if the full distribution should come back as a panel.

## Verification

```
go test ./... -count=1   → exit 0
go vet ./...             → clean
golangci-lint run        → 0 issues
```

Also driven in a browser (`moai web --port 3199`, Playwright): settings, the 20-field MCP tab, the agents table, the profile popover, and `/specs`. Three layout defects only visible on screen were found and fixed — long config keys overflowing the dense label column, the popover rename field crushed to two characters wide, and the close-debt panel buried 31,000px below the table.

Bonus: three pre-existing `errcheck` failures in `events.go` are fixed, since they would have blocked CI.

🗿 MoAI

## Summary by CodeRabbit

* **New Features**
  * Added a redesigned light-mode web console with navigation, settings tabs, profile popovers, localized labels, and improved responsive layouts.
  * Added Overview, Kanban, Monitor, and SPEC catalog screens with searchable details, status summaries, progress tracking, and monitoring data.
  * Added live browser updates for project changes, with connection indicators and polling fallback.
  * Kanban session records now display lead and companion roles.

* **Bug Fixes**
  * Improved settings navigation, validation feedback, accessibility states, and handling of incomplete or unreadable project data.


GoosLab:
@coderabbitai review

PR #1538, opened 2026-08-15
feat(kanban): name the lead session so it survives /clear

GoosLab:
## Problem

A kanban **lead** loses its session name on every `/clear`; companions do not.

Claude Code's documented behavior (`code.claude.com/docs/en/sessions`, `/clear`):

> You keep a name you set with `--name` or `/rename` in the new conversation, but not an AI-generated session title

So an explicit name survives and a generated title does not. Companions are launched from the SessionStart notice (`internal/hook/session_start_kanban.go:119`), which prints `moai cc -k --name <role>-<run-id>` — an explicit name. The lead is launched as a bare `moai cc -k`, so it only ever had a generated title.

`grep -rn 'append.*"--name"' internal/ --include='*.go'` returned no matches before this change: nothing in the codebase supplied a name.

## Change

Both lead branches (`cc.go`, `glm.go`) append `--name lead-<run-id>`, reusing the run id `enterKanbanMode` publishes at `internal/cli/kanban.go` rather than minting a second one.

- `kanban.LeadLabel(runID)` — `<role>-<run-id>` form, same as `CompanionLabel`, but never satisfies `SplitCompanionLabel` because `RoleLead` is absent from `CompanionRoles`. The injected name therefore cannot reclassify a lead as a companion.
- `operatorSuppliedName(args)` — mirrors the existing `operatorSuppliedSettings` precedent, including its `--` pass-through discipline.
- `leadNameArgs(args)` — two self-gates keep the call site unconditional: an operator-supplied name wins, and an absent run id yields no flag rather than the nonsense name `lead-`.

### Operator-supplied `--name` is never overwritten

`parseCompanionLabel` matches only the companion *shape*, so a lead named `board-watch` reads as "no name" there. `operatorSuppliedName` asks the different question — does a name exist at all — across all four forms claude accepts (`--name v`, `--name=v`, `-n v`, `-n=v`), plus a dangling `--name` with no value. A name past `--` is the backend's own argument and is left alone. Covered by `TestOperatorSuppliedName` and `TestLeadNameArgs_NeverOverridesOperatorName`.

`moai glm -k` gets the same treatment — its lead branch mirrors `cc` exactly and has the identical problem.

## Verification

`go build ./...`, `go vet ./...`, `go test ./...` all run. Three test failures reproduce **identically on the unmodified baseline** (measured by stashing this change and re-running), and none of them reach the new code:

- `TestHandleCodexReviewGate_LiveCodexBlocksInjectionAndKey` — live codex verdict
- `TestNavigatorEnrich_AtomicWriteBarrier` — goroutine barrier
- `TestPreTool_AstGrepSkipReasonSurfaces` — `internal/hook`

`gofmt -l` flags `kanban.go` and `glm.go`, both in regions this PR does not touch (pre-existing const alignment / import ordering); left alone per scope discipline.

## Out of scope

t21 (lead env run-id vs session suffix mismatch) is untouched — whether that mismatch comes from relaunch is unobserved.

🗿 MoAI

### DIFF
--- a/internal/cli/cc.go
+++ b/internal/cli/cc.go
@@ -117,6 +117,9 @@ func runCC(cmd *cobra.Command, args []string) error {
 	case kanbanBranchLead:
 		defer enterKanbanMode(specID)()
 		recordKanbanSession(specID, kanban.BackendClaude, kanban.RoleLead)
+		// A lead launched bare has only an AI-generated title, which claude
+		// discards on /clear; naming it explicitly is what survives.
+		filteredArgs = append(filteredArgs, leadNameArgs(filteredArgs)...)
 		settingsFlag, settingsCleanup := prepareKanbanSettings(filteredArgs)
 		if len(settingsFlag) > 0 {
 			filteredArgs = append(filteredArgs, settingsFlag...)
--- a/internal/cli/glm.go
+++ b/internal/cli/glm.go
@@ -185,6 +185,8 @@ func runGLM(cmd *cobra.Command, args []string) error {
 	case kanbanBranchLead:
 		defer enterKanbanMode(specID)()
 		recordKanbanSession(specID, kanban.BackendGLM, kanban.RoleLead)
+		// See cc.go: glm mirrors the lead branch exactly.
+		filteredArgs = append(filteredArgs, leadNameArgs(filteredArgs)...)
 		settingsFlag, settingsCleanup := prepareKanbanSettings(filteredArgs)
 		if len(settingsFlag) > 0 {
 			filteredArgs = append(filteredArgs, settingsFlag...)
--- a/internal/cli/kanban.go
+++ b/internal/cli/kanban.go
@@ -270,6 +270,60 @@ func parseCompanionLabel(args []string) (label string, ok bool) {
 	return "", false
 }
 
+// operatorSuppliedName reports whether the operator named this session
+// themselves, in any of the four forms claude accepts (`--name v`, `--name=v`,
+// `-n v`, `-n=v`), before the pass-through marker.
+//
+// It is deliberately NOT parseCompanionLabel: that function matches the
+// companion SHAPE, so a lead the operator named `board-watch` reads as "no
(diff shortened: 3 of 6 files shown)


GoosLab:
## Merged without a CodeRabbit verdict — alternative evidence recorded

Same gate outcome as #1539: the `CodeRabbit` check row reads `pass` but its detail is `Review rate limited`, so no Merge Risk verdict exists and the hardened gate (#1536) correctly withheld the merge. Merging by hand with the substitute evidence the t26 card prescribes.

**Independent re-measurement by the kanban lead:**
- The safety property that matters most holds structurally, not just by test: `CompanionRoles = []string{"plan", "run", "review", "sync"}` contains no `lead`, so the injected `lead-<run-id>` label can never satisfy `SplitCompanionLabel` and cannot reclassify a lead session as a companion
- `operatorSuppliedName` matches all four forms claude accepts (`--name v`, `--name=v`, `-n v`, `-n=v`), stops at `--`, and treats a dangling `--name` as a declaration — so an operator-chosen name is never overwritten
- The run id is read from `config.EnvMoaiKanbanID` as published by `enterKanbanMode` rather than re-generated, which is why the injection sits after that call. Re-generating would have produced a name whose run id disagreed with the environment — the exact shape of the t21 mismatch
- `glm.go` receives the same treatment as `cc.go`, so `moai glm -k` does not stay asymmetric
- The 3 suite failures were attributed by A/B against a stashed baseline, plus a reachability grep showing the new symbols are unreferenced by those tests

**Why this change exists:** per the official docs, `/clear` keeps a name set with `--name` or `/rename` but discards an AI-generated session title. Companions are launched from the SessionStart notice with `--name <role>-<run-id>`, so they keep their identity; a lead launched as a bare `moai cc -k` had no explicit name and lost it on every clear.

CI: all required checks green. Merges cleanly on top of #1539 (`MERGEABLE` / `CLEAN` re-checked after that landed).

🗿 MoAI

PR #1539, opened 2026-08-15
refactor(cli): remove the GLM team-mode settings writers left dead by #1531

GoosLab:
## What

`enableTeamMode` was removed in #1531. It held the only production call to two functions one layer below it, and both survived the deletion:

- `ensureSettingsLocalJSON`
- `injectGLMEnvForTeam`

Neither is reachable from any user session any more — only from tests. This removes them, and corrects two comments on `glm.go` that still described `injectGLMEnvForTeam` as a live `settings.local.json` wire point on `AC-MTP-032a` reachability grounds. That premise died with the caller.

## Premise re-measured after #1531

The "zero callers" claim was re-checked on `main` at `2b4b34354` (the merge commit of #1531), not carried over from the pre-merge snapshot:

```
$ grep -rn 'ensureSettingsLocalJSON\|injectGLMEnvForTeam' internal/ cmd/ pkg/
```

Every hit outside the two definitions is in a `_test.go` file. Zero non-test callers.

## Two tests kept, and why

Sixteen tests existed only to drive the removed functions and go with them. Two do **not**, and are re-pointed at `mutateSettingsLocal` directly instead:

- `clifix_critical_repro_test.go` — the RMW round-trip guard (unknown top-level keys survive a write)
- `clifix_concurrency_repro_test.go` — the N-goroutine concurrent-writer guard

The contract these guard is the locked + atomic read-modify-write **seam**, not the removed function that happened to call it. `ensureSettingsLocalJSON` was the vehicle, not the subject. Deleting them alongside their vehicle would have retired the M1 concurrency guard silently, which is exactly the failure this PR is cleaning up one instance of. Re-pointing keeps the guard pointed at the thing that is still live.

Flagging this explicitly because a re-pointed test can read as "changed the test until it passed" — the diff should be reviewed against that suspicion, not around it.

## Accepted coverage loss

`TestEnableTeamModeInjectsGLMEnvAndTmux` asserted that a pre-existing unrelated env var survives injection. With the injector gone there is no surface left to assert it against, so that one assertion lapses with no replacement. Nit, since the function it guarded is itself dead.

## Verification

| Check | Result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `go test ./...` | exit 1 — 4 failures, all attributed below |
| `go test ./internal/cli/ -count=1` | 1 failure, pre-existing (see below) |
| `gofmt -l` on touched files | `glm.go` only — already unformatted on `main`; no new defect introduced |

Full-suite failures, each A/B-verified against a clean `2b4b34354` tree in this same worktree rather than assumed:

- `TestHandleCodexReviewGate_LiveCodexBlocksInjectionAndKey` (`internal/cli`) — known pre-existing, fails on clean `main`.
- `TestPreTool_AstGrepSkipReasonSurfaces` (`internal/hook`) — fails on clean `main` in isolation. Pre-existing.
- `TestBranchGuard_Latency` (`internal/hook`) — passes in isolation on both clean and dirty trees; fails only under full-suite load. Load-sensitive latency assertion.
- `TestBuilderNormalizesMode` (`internal/statusline`) — passes in isolation on both clean and dirty trees. Under the full suite it compared `M9` against `M10`: the test reads live `git status`, and `internal/hook`'s perf tests rewrite `.moai/specs/SPEC-HOOK-PRETOOL-PERF-001/{baseline,postchange}.md` in place while it runs, so the modified-file count moves between the test's two snapshots. Cross-package interference, independent of this change.

Those regenerated fixture files were kept out of this commit.

`internal/hook` and `internal/statusline` are untouched here; the diff is confined to `internal/cli`.

🗿 MoAI


GoosLab:
## Merged without a CodeRabbit verdict — alternative evidence recorded

The hardened auto-merge gate (#1536) withheld this merge for the right reason. Run `31864150325`:

```
CodeRabbit status: success
##[notice]No CodeRabbit verdict comment found - withholding merge
```

The `CodeRabbit` check row reads `pass`, but its detail is `Review rate limited` — the review never ran, so there is no Merge Risk verdict to read. This is exactly the failure class the gate was built to stop: absence of review previously presented as review success. The head-freshness guard passed (`TESTED_SHA` = PR head).

Merging by hand with the substitute evidence the t26 card prescribes.

**Independent re-measurement by the kanban lead:**
- Both dead function definitions are gone: `grep -c 'func ensureSettingsLocalJSON\|func injectGLMEnvForTeam'` over the PR's `glm.go` returns `0`
- Pre-flight measurement confirmed zero production callers before deletion; all remaining textual hits are comments plus one negative assertion
- 9 files, +34 −1164
- Two repro tests were re-pointed to the live `mutateSettingsLocal` seam rather than deleted, preserving the lock+atomic-RMW concurrency guard the removed helper merely vehicled
- The 4 suite failures were attributed by A/B against clean `2b4b34354` in the same worktree, not assumed. `TestBuilderNormalizesMode` was traced to cross-package interference: the perf fixtures under `SPEC-HOOK-PRETOOL-PERF-001` are regenerated in place while a statusline test reads real `git status`, moving the modified-file count between snapshots

CI: all required checks green, Race Test included.

**Known follow-up (pre-existing, not from this PR):** the `Skip if not open or already merged` step in `auto-merge.yml` uses `exit 0`, which ends the step but not the job, so every later step still runs. Present since before #1536.

🗿 MoAI

PR #1540, opened 2026-08-15
docs(rules): add cross-session messaging doctrine

GoosLab:
## Summary
- Adds an always-loaded workflow rule (`.claude/rules/moai/workflow/cross-session-messaging.md`, mirrored to the template source) documenting how the orchestrator uses Claude Code's cross-session `ListAgents`/`SendMessage` channel: its three bounding properties, where it sits against the session registry / handoff / agent teams, the HARD boundaries (never route a user decision through a peer, never delegate what this session may not do), role-boundary dispatch vs offloading, and observed addressing/reply frictions.
- No code changes — documentation only.

## Test plan
- [x] Rebased cleanly onto current `origin/main`
- [x] Verified the local `.claude/` copy and the `internal/template/templates/` mirror are byte-identical
- [x] Reviewed content for template neutrality (no SPEC IDs, REQ tokens, internal dates, commit SHAs, or CLAUDE.local references)

🗿 MoAI

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for cross-session messaging between development sessions.
  * Documented message delivery, addressing, availability, configuration, and reply handling.
  * Clarified permission boundaries, concurrency considerations, and appropriate use cases.
  * Added safeguards against routing user decisions or unauthorized work through other sessions.

### DIFF
--- a/.claude/rules/moai/workflow/cross-session-messaging.md
+++ b/.claude/rules/moai/workflow/cross-session-messaging.md
@@ -52,9 +52,17 @@ Worktree isolation remains the structural fix for a write conflict. Messaging sh
 
 Conversely, after landing a change that invalidates what a peer is building on — a schema change, a renamed symbol, a merged branch — notifying the affected peer is appropriate without being asked.
 
-## Session naming
+## Addressing, sending, and replying
 
-A session answers to the name set at launch or by rename; unset, the runtime derives one from the working directory, so parallel sessions in one project collide on a shared prefix and are told apart only by a short identifier. Where a launcher starts a session bound to a known unit of work, passing an explicit name makes peers addressable by what they are doing rather than by where they run.
+A session answers to the name set at launch or by rename; unset, the runtime derives one from the working directory, so parallel sessions in one project collide on a shared prefix and are told apart only by a short reference. Where a launcher starts a session bound to a known unit of work, passing an explicit name makes peers addressable by what they are doing rather than by where they run.
+
+Three frictions are observed in practice and are worth expecting rather than rediscovering:
+
+- **A bare name can be refused.** Sending to a peer by name alone may come back asking for the name plus its short reference before it will resolve. Treat the first refusal as routine: re-send with the reference the error supplies, rather than assuming the peer is unreachable. The user-facing peer listing does not show these references — only the discovery tool's output does — so the reference is read from the tool result or from the refusal itself.
+- **A reply address is not guaranteed to route.** A recipient may be unable to answer the sender it was addressed by and fall back to guessing a peer. Consequently a message must carry enough identification for a human or a peer to route the answer manually: name the sending context and what the answer is for. Never assume a reply will land automatically, and never make the sender's identity implicit.
+- **The sender's permission class is disclosed.** An arriving message states whether its sender bypasses permission prompts, and that disclosure is what the receiver's inbound default keys on. A message from a bypassing sender is more likely to be held for approval, so a session that expects to be answered promptly should not assume delivery.
+
+An arriving message identifies its origin by socket address rather than by name. Where a reply is needed, copy the origin exactly as given rather than re-deriving it from a listing.
 
 ## Configuration surface
 
--- a/.claude/rules/moai/workflow/cross-session-messaging.md
+++ b/.claude/rules/moai/workflow/cross-session-messaging.md
@@ -36,6 +36,10 @@ Messaging complements the registry rather than replacing it: the registry says *
 
 [ZONE:Evolvable] [HARD] **Send facts, not instructions to mutate shared state.** A message may report what landed, what broke, what a decision was, or ask a question. It must not direct a peer to edit configuration, rewrite doctrine, or take a hard-to-reverse action; those remain gated in the receiving session by its own rules and prompts.
 
+[ZONE:Evolvable] **Role-boundary dispatch is permitted; offloading is not.** Where sessions are standing roles in a declared topology — one coordinating session and workers that each own a stage of the pipeline — a coordinating session may dispatch a work item to the session whose role owns that stage, and may ask for its completion status. Three conditions make this dispatch rather than offloading: the target's role is declared in advance rather than chosen because it happened to be idle, the work item is a **pointer into shared source of truth** (an identifier, a path, a contract section) rather than the work itself, and each worker writes to an isolated tree so concurrent workers cannot collide. Absent all three, it is offloading — see the anti-pattern below.
+
+[ZONE:Evolvable] **Do not let a dispatch depend on the reply arriving.** Because reply routing is not guaranteed, completion must also be observable in the shared source of truth — a progress record the coordinator can read — with the message serving as prompt notification rather than as the record. A coordinator that advances only on received replies stalls silently when one is lost.
+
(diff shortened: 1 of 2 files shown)


GoosLab:
*Superseded.* This comment carried an earlier draft of the review-gate record, written before the merge was observed, and a duplicate was created while editing it. The single current record — including the observed merge commit and the fact that no automated review preceded it — is the later "Review-gate record for `d2dc0dc66`" comment on this PR. Read that one; this text is left only so the link does not break.

GoosLab:
## Review-gate record for `d2dc0dc66`

### 1. No re-review ran on these edits

The CodeRabbit check row on this PR reads **`pass`**, and that row is misleading. The commit status on the merged head is:

```
CodeRabbit | state=success | description=Review rate limited
```

It went `pending | Review in progress` first, then settled into rate-limited — so it started and gave up rather than never firing. The only five inline comments on this PR are the **06:03Z originals against the previous head**; zero were posted against `d2dc0dc66`, and the count was still five at merge time. **Nothing reviewed the corrections below.** The green row is not a verdict on them.

It was never going to resolve on its own either: CodeRabbit does not re-review a head merely because the rate limit expires — it needs a new commit or an explicit review trigger, and neither was issued here.

### 2. CI on `d2dc0dc66`

All gating checks passed: Template Neutrality Audit (33s), Constitution Check (1m4s), Lint (2m41s), lsel-leak-guard (29s), Analyze (Go), all five cross-platform builds, Release PR Multi-OS Gate.

One caveat stated rather than glossed: `Test (ubuntu-latest)` finished in **2s**, and Race Test / Integration Tests / the Test matrix reported `skipping`. This is a markdown-only change, so the paths-filter short-circuited the Go test jobs. CI green here means *the jobs that gate a docs change passed* — not that the suite ran. Locally, `MOAI_TEMPLATE_LEAK_STRICT=1 go test ./internal/template/ -count=1` passed in 174.4s, which is the guard that actually covers this change.

### 3. Per-finding verdicts

Each finding was checked against `code.claude.com/docs/en/cross-session-messaging`, fetched directly, before any edit.

**Method note, because it changes how much weight the review carries.** CodeRabbit's own analysis chains show its fetch of that page returned 298 bytes — a failure — after which it reasoned from **web-search summaries** carrying a bare `[1]` citation rather than from the page itself. Its conclusions were therefore checked against the real page, not adopted.

| # | Finding | Verdict |
|---|---------|---------|
| L15 | cross-machine transport misstated | **CONFIRMED** |
| L39 | dispatch guard inverted | **CONFIRMED** |
| L43 | "costs the user nothing" | **CONFIRMED** |
| L69 | socket-origin / addressing | **CONFIRMED IN PART, PARTLY REFUTED** |
| L80 | held-message expiry conflated | **CONFIRMED** |

**L39 — CONFIRMED.** Three conditions are required together, but the guard read *"Absent all three, it is offloading"* — which fires only when every condition is missing, the harmless case, rather than when any one is, the dangerous case. Now reads *"All three must hold together; absent any one of them"*. A defect on its own terms, needing no external evidence.

**L15 — CONFIRMED.** Docs, verbatim: *"Starting a conversation with a session on another of your machines requires Claude Code v2.1.225 or later and a target that appears in the listing. Before v2.1.225, Claude could only reply to a message that arrived from one."* The rule's blanket "reply-only … never open an exchange" was the pre-v2.1.225 behaviour. The genuine narrowings are different ones and are now stated: a send from a session **not itself connected to Remote Control** arrives *"without a reply address, so the receiving Claude can't answer it"*, and a cloud session receives without being able to answer. The same error had spread to two places the review did not flag — the opening scope line and the availability-trap paragraph — and both were corrected, since three places disagreeing would be worse than one being wrong.

**L43 — CONFIRMED.** Docs: *"Once delivered, the message counts toward usage like a prompt you type."* The text now says what is actually saved (a user mediation round-trip) and states plainly that it is not free.

**L69 — CONFIRMED IN PART, AND PARTLY REFUTED.**

*Confirmed:* an arriving message carries **both** a sender name and a reply address — docs: *"Claude receives it with the sender's name and a reply address"* — so *"identifies its origin by socket address rather than by name"* was wrong. Replying by name is the normal path; the address is the fallback.

*Also confirmed:* a bare name resolves whenever *"One session answers to the name"*, and a short identifier is used only when *"several sessions share the name, or Claude Code couldn't check everywhere your sessions run"*. So refusal is the exception, not the routine first outcome the rule described.

*REFUTED and deliberately NOT adopted:* the review also asked that the sentence *"the user-facing peer listing does not show these references — only the discovery tool's output does"* be replaced with a claim that `/list-agents` displays the short identifier. **The page does not say that.** It attributes the short identifier to the assistant's own listing (*"Claude adds a short identifier to each row of its listing"*) and names the **working directory** as what distinguishes same-named sessions in the user-facing `/list-agents` output. Adopting the correction would have replaced a true sentence with a false one, so the original stands.

The same finding also asserted that "direct replies to listed targets require v2.1.225 or later"; the docs place that version gate on *starting a conversation* with an other-machine session, not on replies. The docs' version was written, not the review's.

**L80 — CONFIRMED.** Docs: *"Set `dialogExpiry` to `never` to keep default-held messages until the session ends. A message held by an explicit `hold` setting doesn't expire; Claude Code delivers it only when an `accept` later applies."* The rule said a held message in a non-interactive worker "stays held there", which is pre-v2.1.225 behaviour; the docs now give it the same five-minute deadline with expiry reported to the sender. The unattached-background-terminal asymmetry is also in the docs and is now recorded.

### 4. Independent second read

The coordinating session reports having read all 8 hunks and corroborated two items from live runtime behaviour rather than from the docs page:

- **L69** — cross-session messages arrive carrying both `from="uds:…"` and `from-name="…"`, and replies addressed to the **name alone** have delivered repeatedly across several sessions. Its `ListAgents` output shows rows of the form `<role>-<run-id> [<ref>]`, which is direct evidence that the **discovery tool** does carry the short reference — the sentence the review wanted overwritten.
- **L15's cloud narrowing** — the `ListAgents` tool description states a cloud session *"receives your message but cannot message any session back yet"*.

This is a second human-directed read of the diff. It is **not** a substitute for the automated review, and it is not recorded here as one.

### 5. Merge status — merged unreviewed, after a contested decision

**Merged**: `b8e30c558c7509cd883eaa36b154020378c89797`, 2026-08-15T06:42:52Z, by `GoosLab`, from head `d2dc0dc66`. Observed, not anticipated.

**No fresh automated review preceded it.** At merge time the CodeRabbit status on the merged head still read `Review rate limited`, and the inline-comment count was still the original five against the earlier head. This landed on the strength of the evidence above and the second read in §4 — not on a review.

The decision to do that changed three times within about an hour. The sequence is kept visible rather than flattened to its outcome, because the reversals are the honest shape of what happened and a record showing only the final state would hide that this was contested:

1. **Proposed** — merging on the recorded evidence, as a way past the rate limit.
2. **Withdrawn** — the standing decision became *wait for a real review*, and this PR was held. Nothing was merged while that held.
3. **Granted** — the instruction reversed again to merge all unmerged PRs, this one included, and the merge above followed.

What survives all three positions: the green CodeRabbit row on `d2dc0dc66` means "rate limited", not "reviewed". Anyone auditing this merge later should read it as *merged on recorded evidence without an automated review*, which is what it was — and should not read that row as a passed gate.

PR #1543, opened 2026-08-15
fix(kanban): adopt the lead run id from an operator-supplied session name

GoosLab:
Card **t21**. Root-cause report: `.moai/reports/backlog/t21-kanban-runid-investigation.md`.

## The defect

A lead session carries its run id on **two** surfaces and nothing joined them.

| Surface | Value came from |
|---|---|
| Session identity (`--name lead-<id>`) | the operator's argv |
| Run id (`MOAI_KANBAN_ID`, the leader socket, everything the notice prints) | `NewRunID()` — a fresh mint on every launch |

`enterKanbanMode` minted unconditionally. `leadNameArgs` returned `nil` when the operator supplied a name — moai declined to *add* a name, but never *adopted* the id embedded in the one it was handed. The SessionStart notice composes its companion launch commands from the environment id, so a session named `lead-tjqim9` printed `moai cc -k --name plan-tjr8zx`, and anyone who copied that line opened a companion on a run no lead was listening to.

Measured live on two independent lead processes (report §E5/§E6): both were named with an id 8-9 hours older than the one their environment carried.

The companion branch cannot diverge this way — `enterKanbanCompanionMode` derives its id from its label, *"so the two can never disagree"*. **The asymmetry was the defect.**

## The fix

**F1 — make the lead branch symmetric with the companion branch.**
- `kanban.SplitLeadLabel`, the counterpart of `SplitCompanionLabel`, admitting *exactly* the same run-id shape.
- `enterKanbanMode(specID, leadLabel)` adopts the run id from a `lead-<run-id>` name, minting only when no lead-shaped name is present.
- Both launchers pass it. **`glm.go` needed the same change** — it mirrors `cc.go` exactly, and was measured rather than assumed.
- The four-form `--name` scan is now shared between `parseCompanionLabel` and the new `parseLeadLabel`, so the two shapes cannot drift apart again.

**F2 — assert the environment, not the return value.** The prior suite passed because it checked only that `leadNameArgs` returned `nil`; the published id was never inspected. That is precisely why the divergence shipped green. The tests now assert `MOAI_KANBAN_ID` *and* the leader socket path both carry the id from the operator's name.

**F4 — make a residual disagreement self-announcing.** The lead notice now prints the session name that belongs to the run it is announcing, in all four locales. An operator whose terminal name differs sees it when the notice appears, not when a copied command opens an orphan.

### Shape guard

`SplitLeadLabel` admits what the companion side admits. `lead-notarunid` **is** adopted — `notarunid` is a well-formed run id as far as the grammar is concerned, exactly as `run-notarunid` is accepted as a companion. Rejected (falls back to minting): `board-watch`, `lead-`, `lead-ABC123`, `lead-a-b`. Tightening the lead side alone would reintroduce an asymmetry of its own.

`resolveKanbanBranch` routing is unchanged: `RoleLead` is absent from `CompanionRoles`, so no label can satisfy both discriminators — pinned by a new disjointness test.

## Deliberately NOT in this PR

**F3 — persisting the run id across a lead restart.** A relaunched lead still mints a fresh id and orphans companions launched from the previous notice (report §C4). That is a *separate* defect path, it does not explain the observed symptom, and it reintroduces exactly the state file `NewRunID`'s doc comment avoids. It deserves its own decision, not a ride-along. **Follow-up.**

## Verification

- `go build ./...`, `go vet ./...` — clean
- `golangci-lint run ./internal/cli/... ./internal/kanban/... ./internal/hook/...` — 0 issues
- `go test ./internal/kanban/ -count=1` — ok
- `go test ./internal/cli/ -count=1 -run 'Kanban|LeadName|LeadLabel|OperatorSuppliedName|ParseLeadLabel'` — ok

Two `gofmt -l` hits in the touched files (`internal/cli/kanban.go` const alignment, `internal/cli/glm.go` import order) were confirmed **pre-existing on `origin/main`** and left alone.

🗿 MoAI

### DIFF
--- a/internal/cli/cc.go
+++ b/internal/cli/cc.go
@@ -115,7 +115,13 @@ func runCC(cmd *cobra.Command, args []string) error {
 	label, isCompanion := parseCompanionLabel(filteredArgs)
 	switch resolveKanbanBranch(kanbanEnabled, isCompanion) {
 	case kanbanBranchLead:
-		defer enterKanbanMode(specID)()
+		// An operator-supplied `lead-<run-id>` name carries the run id this
+		// session already belongs to — a relaunch of an existing lead, or a
+		// board whose companions are already open. Adopting it is what keeps
+		// the session name, MOAI_KANBAN_ID, the leader socket path, and the
+		// companion commands the SessionStart notice prints on one run.
+		leadLabel, _ := parseLeadLabel(filteredArgs)
+		defer enterKanbanMode(specID, leadLabel)()
 		recordKanbanSession(specID, kanban.BackendClaude, kanban.RoleLead)
 		// A lead launched bare has only an AI-generated title, which claude
 		// discards on /clear; naming it explicitly is what survives.
--- a/internal/cli/cc_test.go
+++ b/internal/cli/cc_test.go
@@ -547,7 +547,7 @@ func TestEnterKanbanMode_RestoresPriorValue(t *testing.T) {
 	t.Setenv(config.EnvMoaiKanban, "prior-kanban")
 	t.Setenv(config.EnvMoaiKanbanSpec, "SPEC-PRIOR")
 
-	restore := enterKanbanMode("SPEC-PLACEHOLDER")
+	restore := enterKanbanMode("SPEC-PLACEHOLDER", "")
 	if got := os.Getenv(config.EnvMoaiKanban); got != "1" {
 		t.Errorf("inside Kanban Mode %s = %q, want %q", config.EnvMoaiKanban, got, "1")
 	}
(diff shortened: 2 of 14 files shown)


GoosLab:

1. **CodeRabbit status read**: `Review rate limited` — read against head SHA `6ce2174a4ff9a337980ad3776d70bea7910a98b6` via `gh api repos/modu-ai/moai-adk/commits/<sha>/status --jq '[.statuses[] | select(.context == "CodeRabbit")] | max_by(.created_at) | .description'`.

2. **CI green on the same SHA** (`gh pr checks 1543`): Analyze (Go), Build (darwin/amd64, darwin/arm64, linux/amd64, linux/arm64, windows/amd64), CodeQL, Constitution Check, Race Test, Test (ubuntu-latest), Release PR Multi-OS Gate, Integration Tests (ubuntu/macos/windows), Lint — all `pass`.

3. **PR's own verification evidence** (from the PR body): `go build ./...` and `go vet ./...` clean; `golangci-lint run ./internal/cli/... ./internal/kanban/... ./internal/hook/...` — 0 issues; `go test ./internal/kanban/ -count=1` ok; `go test ./internal/cli/ -count=1 -run 'Kanban|LeadName|LeadLabel|OperatorSuppliedName|ParseLeadLabel'` ok. Two pre-existing `gofmt -l` hits confirmed present on unmodified `origin/main` and left untouched.

PR #1544, opened 2026-08-15
fix(hook): carry the permission decision through the handler merge

GoosLab:
## The defect

`registry.Dispatch` pre-seeds its merged output with the event default — already `permissionDecision: "allow"` for PreToolUse under the autonomous permission modes — and `mergeHandlerOutput` copied every `hookSpecificOutput` field **except** `PermissionDecision` / `PermissionDecisionReason`. A handler's `"ask"` was dropped on the floor and the pre-seeded `"allow"` reached the wire.

Only `"deny"` survived, because deny short-circuits (`isBlockDecision`) before the merge runs.

**Blast radius**: every `ask` verdict the PreToolUse hook can produce — 27 `AskPatterns` file patterns (`package.json`, `Dockerfile`, `.github/workflows/*`, `terraform/*.tf`, `k8s/*`, `Jenkinsfile`, lock files, …) and 10 `AskBashPatterns` commands (`git reset --hard`, `git clean -fd`, force-push, `prisma migrate reset`, …). All documented as requiring confirmation; none actually required it.

The omission predates the `mergeHandlerOutput` refactor that broadened the merge without adding the permission fields — a long-standing latent defect, not a recent regression.

## The fix

A **precedence ladder** (`deny > ask > allow > unset`), not a field copy. A blind copy would be wrong in the other direction: it would let a later handler's `allow` clobber an earlier handler's `ask` or `deny`. The reason travels with the decision — an `ask` with an empty `permissionDecisionReason` shows a dialog that explains nothing.

The sibling `PermissionRequest` `decision.behavior` field had the same omission. It was **not** a live defect: `defaultOutputForEvent` returns a bare `&HookOutput{}` there, so the nil-`HookSpecificOutput` early-assign branch carried the object through, and only one handler is registered for the event — sound by accident on two counts. It now follows the same ladder explicitly, with regression cases.

## The tests are the real fix

This shipped green because **every** `ask` assertion called `preToolHandler.Handle` directly and never crossed `registry.Dispatch`: the producer was tested, the consumer was not. `deny` had both handler-level and end-to-end coverage, which is exactly why `deny` stayed correct while `ask` rotted unnoticed.

Added, and each verified to **fail** with the fix removed:

1. `internal/hook` registry-level — `Dispatch(EventPreToolUse, …)` with an `AskPatterns` file, asserting `"ask"` **and** a non-empty reason
2. `internal/hook` registry-level — the same for `git reset --hard` (`AskBashPatterns`)
3. `internal/hook` — a 7-case precedence table pinning the ladder in both directions
4. `internal/hook` — `PermissionRequest` `decision.behavior` cases
5. `internal/cli` — wire-format assertions on the JSON reaching stdout, pinning `ask` end-to-end the way `deny` already is

A deny-still-short-circuits guard is included so the path that was already correct stays that way.

## ⚠️ Behaviour change users will notice

**This will start surfacing permission dialogs that have been silently suppressed for a long time.** Editing `package.json`, a `Dockerfile`, a workflow file, or running `git reset --hard` will now prompt for confirmation where it previously proceeded without asking.

That is the correct, documented behaviour and the whole point of the fix — but it is a visible change, it will look sudden, and it may be mistaken for a new bug. Under `defaultMode: bypassPermissions` the difference is starkest: an unattended agent could previously force-push or hard-reset with no gate at all.

## Verification

- `go test ./internal/hook/ ./internal/cli/ -count=1` — green apart from two failures reproduced identically on unmodified `main` (`TestPreTool_AstGrepSkipReasonSurfaces`, `TestHandleCodexReviewGate_LiveCodexBlocksInjectionAndKey`)
- `go vet ./...` — clean; `golangci-lint run` on both packages — 0 issues
- Probes re-run against a freshly built binary: `ask` now reaches the wire with its reason for both the file and Bash paths; `deny` unchanged; ordinary edits and commands still `allow`

🗿 MoAI

### DIFF
--- a/internal/cli/hook_pretool_ask_wire_test.go
+++ b/internal/cli/hook_pretool_ask_wire_test.go
@@ -1,0 +1,145 @@
+// End-to-end wire-format regression test for the PreToolUse "ask" verdict.
+//
+// "deny" already had CLI-level coverage (hook_protocol_fix_test.go); "ask" had
+// none, which is why the registry merge could drop it and the suite stayed
+// green for the whole life of the defect. This test pins "ask" the same way
+// "deny" is pinned: real protocol, real registry, real handler, assertion on
+// the JSON that actually reaches stdout.
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// runPreToolHookCLI feeds payload to the `moai hook pre-tool` subcommand with
+// the real wiring and returns the JSON object written to stdout.
+func runPreToolHookCLI(t *testing.T, payload string) map[string]any {
+	t.Helper()
+
+	origDeps := deps
+	t.Cleanup(func() { deps = origDeps })
+	InitDependencies()
+
+	swapStdinString(t, payload)
+
+	origStdout := os.Stdout
+	rOut, wOut, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("os.Pipe: %v", err)
+	}
+	os.Stdout = wOut
+
+	var runErr error
+	found := false
+	for _, cmd := range hookCmd.Commands() {
+		if cmd.Name() == "pre-tool" {
+			found = true
+			cmd.SetContext(context.Background())
+			runErr = cmd.RunE(cmd, []string{})
+			break
+		}
+	}
+	_ = wOut.Close()
+	os.Stdout = origStdout
+
+	if !found {
+		t.Fatal("pre-tool subcommand not found")
+	}
+	if runErr != nil {
+		t.Fatalf("RunE returned error: %v", runErr)
+	}
+
+	out, err := io.ReadAll(rOut)
+	if err != nil {
+		t.Fatalf("read captured stdout: %v", err)
+	}
+
+	var decoded map[string]any
+	if err := json.NewDecoder(strings.NewReader(strings.TrimSpace(string(out)))).Decode(&decoded); err != nil {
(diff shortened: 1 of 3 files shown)


GoosLab:

1. **CodeRabbit status read**: `Review rate limited` (checked repeatedly over ~15 minutes, no change) — read against head SHA `9cda80e10fa1540582c3427e06d1f00a5ca6b4d1` via `gh api repos/modu-ai/moai-adk/commits/<sha>/status --jq '[.statuses[] | select(.context == "CodeRabbit")] | max_by(.created_at) | .description'`.

2. **CI green on the same SHA** (`gh pr checks 1544`): Analyze (Go), Build (darwin/amd64, darwin/arm64, linux/amd64, linux/arm64, windows/amd64), CodeQL, Constitution Check, Race Test, Release PR Multi-OS Gate, Integration Tests (ubuntu/macos/windows), Lint, Test (ubuntu-latest) — all `pass`.

3. **PR's own verification evidence** (from the PR body): `go test ./internal/hook/ ./internal/cli/ -count=1` green apart from two pre-existing failures reproduced identically on unmodified `main`; `go vet ./...` clean; `golangci-lint run` 0 issues on both packages; probes re-run against a freshly built binary showing `ask` now reaches the wire with its reason for both the file and Bash paths, `deny` unchanged, and ordinary edits/commands still `allow`. Five new regression tests added, each confirmed to fail with the fix removed (registry-level ask+reason for file and Bash patterns, a 7-case precedence-ladder table, `PermissionRequest.decision.behavior` cases, and wire-format JSON assertions).

PR #1545, opened 2026-08-15
docs(docs-site): rewrite web console docs for the operations shell and close v3.1 gate items

GoosLab:
Cards **t35** and **t32** (items 1, 2, 5, 6, 7 only). Items 3, 4 and 8 are deferred by operator decision and are untouched.

## t35 — the web console docs described a console that no longer exists

Eight pages described a 9-tab console. The surface since #1523 is a five-area operations shell. Measured coverage of the new surface across those files before this change: Kanban 0, Monitor 0, `/specs` 0, `/monitor` 0, `/kanban` 0, SSE 0.

Rewritten against the implementation — `internal/web/shell.templ`, `screens.templ`, `viewmodel_ops.go`, `events.go`, `app.go`, `assets/app.js` — not against the old prose.

Routes verified against a live server (`moai web --port 39511 --no-open --no-reuse`, binary built from this branch):

```
/          200      /kanban   200      /specs   200
/monitor   200      /settings 200      /events  200 (text/event-stream, curl exit 28 = held open)
```

Two corrections to the investigation report's framing, both measured rather than assumed:

- The appbar context chips (`lang`/`model`/`effort`/`dev`) render on **`/settings` only**, not on the ops screens. `settingsCtxChips` is called only from `settings_shell.go`; `screens.go` `shellVM` sets no `Ctx`. Confirmed: `curl / | grep -c chip-kv` → 0, `curl /settings` → the four keys. The ops screens carry the live indicator instead. Documented as measured.
- The four stale screenshots are **dropped** rather than replaced. They show the green `#3d7d5f` primary, the mascot and the horizontal 9-tab strip — none of which exist now. A misleading image is worse than no image; I cannot capture replacements from this environment. Flagging so a follow-up can add fresh captures.

## t32 item 1 — sidebar entries

Added the 7 missing `/advanced/*` entries with 4-locale name maps, titles taken from each page's own frontmatter. Five of them are `new_items` NEW-badge targets — verified in `layouts/partials/menu.html` that the badge renders only inside `range .sub` (lines 76-102), so a page absent from `main.yaml` has no `<li>` and its badge could not render at all.

Confirmed no `menu.html` icon case is needed rather than relying on the report: `.icon` is read only at section level (line 48, inside `<summary>`); sub-items never read it.

## t32 item 2 — `_meta.yaml` congruence

Backfilled 8 keys in en/ja/zh and 2 in ko. All four sets are now 32 keys with **identical key sets** (verified by `diff` of sorted key lists, not just by count).

## t32 item 5 — `workflow.branch_guard.enabled`

Read from source rather than inferred: `internal/config/types.go:562` (`BranchGuardConfig{Enabled bool}`), `defaults.go:660` (`Enabled: false`). Also confirmed the key is **absent** from `internal/template/templates/.moai/config/sections/workflow.yaml`, so distributed users get the compiled default — writing the key is itself the opt-in.

## t32 item 6 — `cli-reference/epic.md` ×4

One correction: the audit lists `--locale` as a flag. It is **not registered**. `epic.go:61-63` declares the variable then discards it with `_ = locale`; only `--json`, `--design-report`, `--marker`, `--base-dir` reach `cmd.Flags()`. Verified with `moai epic status --help`. The page documents four flags.

All example output is real captured output from `moai epic status KANBAN` and `... DESIGN-DOCS` (human and `--json`), not invented.

## t32 item 7 — the parity gate, and why it ratchets

**Failure condition chosen: fail on NEW divergence only, against a checked-in baseline** (`docs-site/.locale-parity-baseline`, 63 pages).

The old check summed `^## ` across each locale tree and compared totals, with no stated pass criterion. Per-page divergences in opposite directions cancel, which is why 63 divergent pages presented as a healthy-looking total.

I rejected the absolute per-page check. It would fail on all 63 pages on its first run, and a gate that fails on day one is disabled within a day — strictly worse than the weak check it replaces. I also rejected a glob allowlist: it would hide *new* divergence inside an already-listed directory. A per-page baseline fails the moment a previously-balanced page loses parity, makes the existing debt explicit and auditable, and can only shrink — the gate also prints converged pages so the baseline gets pruned.

Held-in evidence (the gate actually catches what it claims): injected one extra `## ` into `en/advanced/moai-web-console.md` and re-ran — the gate printed `advanced/moai-web-console.md` as new divergence. Probe reverted; gate returns to clean.

Note this is a *check recipe* in a user-owned harness skill, not CI wiring — it runs when the oss-docs harness verifies.

## Verification

| Check | Result |
|---|---|
| `hugo --minify --gc` | exit 0, **no WARN/ERROR**, 24961 ms |
| sitemap | present |
| URL blacklist | no matches |
| Mermaid LR/RL | no matches (all new diagrams are `flowchart TD`) |
| File-existence parity | clean — 144 pages × 4, sorted lists `diff`-identical |
| Per-page section parity | **0 new divergence**, 63 baselined |
| 8 rewritten pages | 12 H2 / 15 H2+ (console) and 7 H2 (web.md), identical in all four locales |
| `_meta.yaml` key sets | identical across ko/en/ja/zh |
| Body-emoji scan | new hits are verbatim CLI output inside ` ```bash ` fences only |

**Not run:** the full Go suite. Per the lead's mid-task instruction (box at load average 358), and because this branch touches no Go and no template mirror. Verified the mirror premise rather than assuming it: `internal/template/templates/.claude/skills/` contains only `moai-*` skills — the `hns-oss-docs-*` harness is user-owned and has no mirror, so Template-First does not apply and `make build` / `catalog.yaml` are not involved. CI runs the suite on this head.

**Pre-existing, not addressed (out of card scope):** `README.ko.md` has 10 H2 against 13 in the other three. No README was touched by this branch.

🗿 MoAI


GoosLab:

1. **CodeRabbit status read**: `Review rate limited` — read against head SHA `87e35fa0aa3914807173227ba7b1cf7db11dca5a` via `gh api repos/modu-ai/moai-adk/commits/<sha>/status --jq '[.statuses[] | select(.context == "CodeRabbit")] | max_by(.created_at) | .description'`.

2. **CI green on the same SHA** (`gh pr checks 1545`): Analyze (Go), Build (darwin/amd64, darwin/arm64, linux/amd64, linux/arm64, windows/amd64), Constitution Check, Lint, Release PR Multi-OS Gate, Test (ubuntu-latest), Validate 4-locale parity — all `pass`.

3. **PR's own verification evidence** (from the PR body): `hugo --minify --gc` exit 0, no WARN/ERROR (24961ms); sitemap present; URL blacklist no matches; Mermaid all `flowchart TD`; file-existence parity clean (144 pages x4, sorted-list diff-identical); per-page section parity 0 new divergence against the 63-page baseline; the 8 rewritten pages have identical H2 structure across all four locales; `_meta.yaml` key sets identical across ko/en/ja/zh; body-emoji scan hits confined to bash-fenced CLI output. Routes verified against a live server built from this branch (/, /kanban, /specs, /monitor, /settings all 200; /events 200 text/event-stream). Full Go suite explicitly not run per the lead's load directive — branch touches no Go/template mirror, verified rather than assumed (only `moai-*` skills are template-mirrored; the touched `hns-oss-docs-*` harness is user-owned, no mirror).

PR #1548, opened 2026-08-15
fix(hook): scope branch-guard scan to the command; correct the exemption doctrine

GoosLab:
Card t34. Two findings, root-caused separately. One turned out to be a code defect; the other turned out to be a documentation defect with working code behind it.

## Finding 1 — the exemptions are unreachable, not broken (rule fix, no code change)

Probing the hook directly separates the axes:

| Probe | Payload / env | Result |
|---|---|---|
| baseline | neither axis | `deny` — `BRANCH_GUARD_VIOLATION: git branch` |
| identity axis | `"agent_type":"manager-git"` in payload | `allow` |
| env axis | `MOAI_BRANCH_GUARD_EXEMPT=1` in the hook process env | `allow` |

Both axes fire on their own. The code is correct; delivery is what fails, and it fails for a different reason per axis:

- **`AgentType`** arrives in the hook payload. Claude Code populates `agent_type` for a main-thread `claude --agent <name>` launch; a subagent spawned through the Agent tool sends none on PreToolUse. In `internal/hook/types.go` the field sits with the SessionStart group and is documented as "custom agent name if --agent flag used".
- **The env sentinel** is read from the hook process's own environment. The hook is a separate process spawned *before* the guarded command runs, so an `export` of it inside that command sets it in a process the hook never sees.

That combination explains the reported symptom exactly: a manager-git subagent exported the sentinel, supplied neither reachable value, and was denied. It then worked around the deny with `git -C <worktree>`, which succeeds because a worktree's git-dir differs from its git-common-dir — the discriminant working as designed, not a second bug.

So the honest fix is the rule, which described exemptions a subagent cannot use. It now states each axis's reachability and names the two routes that do work: run the work in a worktree, or launch the session with the sentinel already in its environment. A unit test pins each axis independently, so a future failure is attributable to delivery rather than to an unimplemented axis.

## Finding 2 — the scan was not scoped to the command (code fix)

`moai todo add "… git switch …"` was denied: the command being run was `moai todo add`, and the match came from data inside a quoted argument that would never execute.

Quoted spans now collapse to a placeholder before pattern matching. **The placeholder is a word, not whitespace, and that is load-bearing** — the first attempt blanked the span and silently un-guarded every branch-state command with a quoted branch name, because `git checkout -b "feat/x"` lost the operand the pattern requires after `-b`. The regression test caught it before it shipped; the falsification arm of that test is what pins it.

End-to-end against the reinstalled binary:

| Command | Before | After |
|---|---|---|
| `moai todo add "… git switch main …"` | deny | **allow** |
| `git commit -m "revert the git switch main change"` | deny | **allow** |
| `git switch main` | deny | deny |
| `git checkout -b "feat/quoted"` | deny | deny |
| `git switch main` with cwd `/tmp` (non-git) | allow | allow (fail-open) |

Residual, accepted and documented: a git invocation hidden inside a shell wrapper no longer matches, since it lives inside the discarded span. For a fail-open, opt-in guard, under-matching a deliberately obfuscated form is the correct direction to err, and the Claude Code worktree isolation guard refuses that shape independently.

## Invariants held

- Fail-open unchanged — deny still requires primary checkout **and** a pattern match **and** non-exempt; every uncertainty path still allows with a stderr advisory and an audit-log line (the non-git-cwd row above exercises it).
- Opt-in default unchanged — `Workflow.BranchGuard.Enabled` still ships false, and the guard is inert for anyone who never opted in. This change touches only the matching step *inside* the enabled path.

## Verification

Reduced profile per the lead's load directive — targeted tests, scoped vet/lint, CI as the full baseline.

- `go test ./internal/hook/ -run 'BranchGuard|BranchState|SubstituteQuoted|IsExemptAgent|CheckBranchState' -count=1` → `ok 28.209s`
- `go vet ./internal/hook/` → clean; `golangci-lint run ./internal/hook/...` → `0 issues.`
- `gofmt` clean on both changed Go files
- Binary reinstalled with the clean rm-then-cp form, `moai version` exit 0; the table above is from that binary
- **Not run**: the full `go test` sweep — the machine is at load average ~370 and four lanes running full suites is the cause. CI is the baseline. Recorded as a gap, not a pass.
- One observation for the load card: `TestBranchGuard_Latency` failed in a batch run and passed alone, with no code change between the two.

Two "command-injection (high)" advisories fired during editing. Both point at prose lines in the doc comment describing process ordering. This file spawns no shell.

🗿 MoAI

### DIFF
--- a/.claude/rules/moai/workflow/main-checkout-branch-guard.md
+++ b/.claude/rules/moai/workflow/main-checkout-branch-guard.md
@@ -130,9 +130,36 @@ flows. The hook applies the doctrine conditionally.
   cwd-normalized `--git-common-dir`.
 - **Exemption mechanism**: the deny is suppressed when EITHER the invoking
   agent identity is the trusted git agent (`HookInput.AgentType ==
-  "manager-git"`, populated on main-thread `claude --agent manager-git`) OR
-  the sentinel environment variable `MOAI_BRANCH_GUARD_EXEMPT=1` is set by
-  the orchestrator when spawning that agent for Late-Branch closure.
+  "manager-git"`) OR the sentinel environment variable
+  `MOAI_BRANCH_GUARD_EXEMPT=1` is present. Both axes are implemented and each
+  fires on its own — but each is read from a different place, and **neither is
+  reachable from inside a tool-spawned subagent**:
+  - `AgentType` arrives in the hook payload, and Claude Code populates
+    `agent_type` for a main-thread `claude --agent manager-git` launch. A
+    subagent spawned through the Agent tool sends no `agent_type` on
+    PreToolUse, so the identity axis cannot fire for it.
+  - The sentinel is read from the hook process's own environment. The hook
+    runs as a separate process spawned **before** the guarded command executes,
+    so an `export MOAI_BRANCH_GUARD_EXEMPT=1` inside that command never reaches
+    it. The variable must be present in the environment Claude Code itself was
+    launched with.
+
+  Exporting the sentinel inside the command being guarded is therefore a no-op.
+  A `manager-git` subagent that needs to mutate branch state has two working
+  routes: do the work in a worktree (`git -C <worktree>`, which the discriminant
+  correctly classifies as non-primary), or have the operator launch the session
+  with the sentinel already in its environment. Reading a `BRANCH_GUARD_VIOLATION`
+  as "the exemption is broken" is a misdiagnosis — the axes work; the values were
+  never delivered.
+
+- **Scan scope**: the pattern set is matched against the command with quoted
+  spans collapsed to a placeholder word, so a match reflects the command being
+  invoked rather than text carried as data. `moai todo add "… git switch …"` is
+  allowed because the command being run is `moai todo add`; `git switch main`
+  and `git checkout -b "feat/x"` both still deny, the latter because the
+  placeholder preserves the operand after `-b`. A git invocation hidden inside a
+  shell wrapper (`bash -c "git switch main"`) is not matched — under-matching an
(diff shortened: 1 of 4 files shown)


GoosLab:

1. **CodeRabbit status read**: `Review rate limited` — read against head SHA `eae65ba4d368cfc6790ce17249afe93413a9a9d4` via `gh api repos/modu-ai/moai-adk/commits/<sha>/status --jq '[.statuses[] | select(.context == "CodeRabbit")] | max_by(.created_at) | .description'`.

2. **CI green on the same SHA** (`gh pr checks 1548`): Analyze (Go), Build (darwin/amd64, darwin/arm64, linux/amd64, linux/arm64, windows/amd64), CodeQL, Constitution Check, Race Test, Test (ubuntu-latest), Release PR Multi-OS Gate, Integration Tests (ubuntu/macos/windows), Lint, Template Neutrality Audit, lsel-leak-guard — all `pass`.

3. **PR's own verification evidence** (from the PR body): `go test ./internal/hook/ -run 'BranchGuard|BranchState|SubstituteQuoted|IsExemptAgent|CheckBranchState' -count=1` ok (28.2s); `go vet ./internal/hook/` clean; `golangci-lint run ./internal/hook/...` 0 issues; `gofmt` clean on both changed Go files; binary reinstalled clean rm-then-cp, `moai version` exit 0, and the before/after command-scan table in the PR body was produced against that reinstalled binary. Explicitly flagged gap in the PR body: the full `go test` sweep was NOT run due to machine load (~370), CI cited as the baseline instead — noting this gap rather than treating CI-only as an unqualified pass.

PR #1542, opened 2026-08-15
docs(kanban): CodeRabbit read discipline, card classes, env-isolated verify form

GoosLab:
Three kanban-dispatch cards land in one file, so they land in one PR — separate PRs would collide on `.claude/rules/moai/workflow/kanban-dispatch.md`.

## What changed

**CodeRabbit is not read from `gh pr checks`** (new subsection under "Completion is read, never trusted"). The check row prints `pass` whether or not a review ran: the commit status `state` is `success` in both cases, and only the description separates `Review completed` from `Review rate limited`. The lead now needs two conditions before counting a CodeRabbit row as evidence — the description reads `Review completed`, and a `Merge Risk:` line exists whose `up to` prefix matches the current `headRefOid`. The rule states plainly that branch protection is not the lever, because the state is `success` in precisely the failing case.

**Card classes and WIP** (new section after backlog entry). Class A direct-close, Class B defect-with-unknown-cause, Class C design change. The justification recorded for a Class A card is the three properties that make skipping columns safe — one file, no design judgement, CI catches the regression — never "it is faster". WIP ceiling is one card per column and only when each occupies a different worktree; for Class A, four sessions each holding a whole card beats pipelining one card through four columns.

**The env-isolated verification form** (new subsection under worktree isolation). `env -u VAR <cmd>` is refused by the agent worktree isolation guard in the Claude Code binary, through an argument-boundary misparse: with `argv[0] == env` the guard scans the whole remaining argv as env's flags, so the inner command's `-run` is blamed on env. The working form is one compound `unset … && <cmd>`, documented with both caveats that make it fragile to "simplification" — it must be a single invocation, and it must not be wrapped in a subshell. The script-file workaround is explicitly rejected as a standing pattern: the guard cannot see inside a script, so it bypasses the git-redirect checks that are the guard's actual purpose.

## Template-First

The mirror under `internal/template/templates/` received the same delta re-applied by hand, not a copy. The added prose carries no internal references, so both trees stay byte-identical.

## Verification

- `MOAI_TEMPLATE_LEAK_STRICT=1 go test ./internal/template/ -count=1` → `ok` (166.6s)
- `go test ./... -count=1` → one failure, `TestHandleCodexReviewGate_LiveCodexBlocksInjectionAndKey`, which calls the live codex backend and whose own failure note says a non-BLOCK may mean codex returned an inconclusive verdict. This diff is two markdown files and cannot reach that path. Not re-baselined against `main` — recorded as a gap rather than claimed as pre-existing.
[...]

PR #1549, opened 2026-08-15
docs(kanban): dispatch in the operator's language; keep verification load lane-local

GoosLab:
Cards t19 and t37. Both land in `.claude/rules/moai/workflow/kanban-dispatch.md`, so one branch and one PR.

> **Stacked on #1542 — merge after it.** This branch is based on #1542's head (`3fd1f6f08`), not on `main`, because #1542 adds three sections to the same file and is still open. The diff here is the two new sections only; the base comparison will look right once #1542 lands.

## t19 — dispatch language: the carve-out holds, and the constitution stays untouched

The card asked for an analysis before an edit, so here is the analysis and the verdict.

**Verdict: the carve-out holds. It is a classification, not a repeal, and it required no change to either language rule.**

**`agent-common-protocol.md` § Language Handling does not oppose it.** Its opening HARD line is the opposite of an English default:

> All agents receive and respond in user's configured conversation_language.

What that section fixes to English is code examples, skill names, technical identifiers, and function/variable/class names — addresses, not prose. So there was nothing to repeal here. The only real gap is that the enumeration never named cross-session messages, so this PR adds exactly one row for that case. That is the single line the card permitted, and it is genuinely needed: without it a dispatch matches none of the listed categories.

**`moai-constitution.md` § Response Language is the only clause that could bind**, via:

> Internal agent communication uses English

But the axis that clause sits on is defined one line above it — *"All user-facing responses MUST be in the user's conversation_language"* — with the three bullets reading as a pair: respond to the user in their language, and use English for the traffic they do not see. The distinction being drawn is **user-facing vs not-user-facing**, not "session-to-session vs in-session". A dispatch the operator watches scroll past is user-facing by the constitution's own criterion.

So classifying a human-observed dispatch as `conversation_language` **applies** that rule rather than contradicting it. No edit to `moai-constitution.md`, which is what the blast-radius constraint wanted anyway. (`CLAUDE.md` §9 carries the same "internal agent comms always English" phrasing on the same axis, and is unaffected for the same reason.)

**The boundary is who reads it.** An `Agent()` subagent prompt reaches no human and stays English. That line is written into the rule so the carve-out cannot be widened by analogy later.

**What stays verbatim in every language**: SPEC IDs, command names and their flags, file paths, session names, and technical identifiers — addresses, and a translated address does not resolve.

## t37 — verification load discipline

Written down with the numbers, since this is the card the incident earned.

Measured: load average reached 413; a neighbouring workspace's build took two and a half minutes and its browser tests timed out. Two causes, both recorded:

1. **Four lanes each running the full suite concurrently**, where a full suite takes five to ten minutes. The rule is now that lane-local verification is scoped to the card, with CI as the *stronger* evidence rather than the fallback — CI runs the full suite in a clean environment against the actual pull-request head, while a full-suite run on a loaded box measures the box. A test that fails in a crowded batch and passes alone has reported contention, not a defect.
2. **A verification recipe that spawned eight spin loops** and put its kill line *after* the long test command. The agent ended before reaching it and twelve spinners ran orphaned for thirty-seven minutes. Load a verification genuinely needs must be cleanup-guaranteed — cleanup-hook-registered kills or a `timeout` wrapper — never a trailing `kill` the process may never reach. A recipe that spawns processes is reviewed as a hazard in its own right.

**Attribution is explicit in the rule**: the fault belongs to the dispatcher who wrote and approved the recipe, not to the lane that ran it as given. A rule that blames the executor teaches the wrong actor to be careful.

Also folded in the finding that explains why this presented as test interference: a failing test left an unbounded spin-loop goroutine burning a core for the rest of that package's run. Load makes tests fail; failing tests can generate load.

## Template-First

Mirror delta re-applied by hand to both files, not copied. The added prose carries no internal references (no PR numbers, dates, SPEC IDs, or SHAs), so both trees stay byte-identical.

## Verification

Per the load directive — the strict template test only, no full sweep, no repeated timing runs.

- `MOAI_TEMPLATE_LEAK_STRICT=1 go test ./internal/template/ -count=1` → `ok 67.635s`
- `diff` of each rule against its mirror → identical
- `make build` regenerated `catalog.yaml` with no content change (rules are not a catalog input, so nothing to stage)

No Go code changed, so no vet or lint run was applicable.

🗿 MoAI

## Summary by CodeRabbit

- **Documentation**
  - Clarified cross-session communication language requirements while preserving technical identifiers.
  - Added guidance for detecting and handling stale or indeterminate sessions.
  - Documented pre-edit synchronization safeguards and their rationale.
  - Added lane-local verification expectations, including scoped testing, cleanup requirements, and CI-based full-suite validation.
  - Clarified that internal subagent prompts remain in English.

### DIFF
--- a/.claude/rules/moai/core/agent-common-protocol-reference.md
+++ b/.claude/rules/moai/core/agent-common-protocol-reference.md
@@ -154,3 +154,33 @@ turn where deferred tools may be needed. See
 - `.claude/rules/moai/workflow/cache-aware-execution.md` — prompt-cache-aware
   ordering (stagger-spawn for parallel same-type agents, gate placement,
   session-loaded file edit timing).
+
+---
+
+## Pre-Edit Sync Check — rationale and enforcement record
+
+> Relocated verbatim from `agent-common-protocol.md` § Pre-Edit Sync Check (Direct-Edit Race Mitigation) to keep the always-loaded file within its size budget. The binding gate (TRIGGER / CHECK / DECIDE / RE-CHECK) and the sweep prohibition remain inline there; this section carries the incident record, the failure analysis of the previous version, and the enforcement-placement assessment.
+
+### Incident record (why the rule is binding)
+
+On 2026-08-15, five Claude sessions worked the same shared primary checkout simultaneously. 104 uncommitted files accumulated there; **73 of them existed on no branch at all** and had to be rescued into PR #1526 by hand. The destructive mechanism is ordinary: a concurrent `git add -A && commit` in one session sweeps another session's uncommitted work into a commit that was never meant to carry it; a branch switch or stash strands it outright. A rule without the incident attached gets skipped; this one has the receipts.
+
+### Why the previous version of this check failed
+
+Two distinct failure modes — they need different fixes:
+
+1. **Nothing executed it.** The check was procedural prose in this rule file. No hook fires it, no tool result reminds the agent of it, and a "run three probes before your first edit" instruction competes with task momentum — and loses. This was the dominant failure on 2026-08-15: the procedure was directionally correct and was simply never run.
+2. **It sampled at the wrong moment.** Even perfectly executed, a once-per-task check cannot see a session that starts mid-task, and the destruction mechanism is not the edit — it is the OTHER session's `git add -A` at commit time. The previous text placed all discipline on the editor and none on the sweeper.
+
+The rewrite addresses both: the decision procedure is compressed into a moment-of-edit gate an agent can actually run, and a sweep prohibition binds the commit-side primitive that does the damage.
+
+### Enforcement assessment — why there is no PreToolUse-on-Edit/Write advisory hook
+
+A per-edit advisory hook was considered and declined (2026-08-15):
+
+- **Measured cost**: the branch-guard PreToolUse hook — the closest comparable, it spawns git subprocesses — averages 135-256ms per invocation. `Edit`/`Write` is the most frequent mutation in a coding session; a per-edit advisory multiplies that tax across every edit of every concurrent session — the exact multi-session load pattern that pushed the box over during the incident day.
+- **Wrong sampling rate**: the probe's answer changes on the scale of minutes (sessions start and stop), not edits. Paying per-edit for a per-task answer taxes the wrong frequency.
+- **Wrong target**: the edit is not the destructive primitive; the sweep is. If mechanical enforcement is ever wanted, the cheap and mechanism-adjacent placement is the commit-time Bash surface — extend the branch-guard hook family to warn on `git add -A` / `git add .` / `git commit -a` in the primary checkout (rare, once per task) — not on every Edit/Write.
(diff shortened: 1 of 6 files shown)


GoosLab:
## Merge record — merged without a CodeRabbit review

**1. CodeRabbit verdict, verbatim.** Head `468820a9a571a17a94f883c269eefc78744395e4`:
`Review skipped: draft pull request`. The PR carried a lead review gate in its body and stayed
draft, so no automated review ran on this head. The `gh pr checks` row nevertheless reads `pass` —
that row is not a verdict, which is the distinction this PR's own sibling (#1542) documents.

**2. CI on that same SHA.** Terminal green, zero failing and zero pending: Build ×5 (linux/amd64,
linux/arm64, darwin/amd64, darwin/arm64, windows/amd64), Lint, Constitution Check, Template
Neutrality Audit, lsel-leak-guard, spec-status-sync, Analyze (Go), Detect (paths-filter),
Release PR Multi-OS Gate, Test (ubuntu-latest).

Which jobs actually RAN matters here: `Test (ubuntu-latest)` completed in 3s and Race Test /
Integration Tests are `skipping`, because paths-filter short-circuits the Go jobs on a docs-only
diff. So "CI green" on this PR means the docs-gating jobs passed — **not** that the Go suite
executed. Recorded explicitly so a later reader does not mistake one for the other.

**3. This PR's own verification, and the reason it is more than housekeeping.**

It removes a live budget breach on `main`. Measured per commit:

| Commit | `agent-common-protocol.md` |
|---|---|
| before #1550 | 39,376 |
| `c8ab130e0` (#1550, merged today) | **43,096 — over the 40,000 budget** |
| `468820a9a` (this PR) | **39,455** |

`main` has been in breach since #1550 landed. That file is always-loaded in full for every agent on
every session, so the overage is paid on every spawn. The hook only fired when a worktree touched
the file; nothing surfaced it at merge time.

The fix relocates weight rather than trimming it: the Pre-Edit Sync Check incident record, the
why-the-previous-version-failed analysis, and the enforcement-placement assessment (~3,000 chars)
move into the path-scoped `agent-common-protocol-reference.md` sidecar, and the Recovery-Signal
Carve-Out render surface is compressed against its SSOT in `runtime-recovery-doctrine.md` §4
(~1,400 chars). Binding content stays inline: the TRIGGER / CHECK / DECIDE tables, the
stale-registry caveat, RE-CHECK, and the sweep prohibition. Both the real file and its template
mirror are byte-identical at 39,455, leaving 545 chars of headroom — real weight removed, not
shaved to fit.

The PR's own additions (the dispatch-language section and the verification-load section) were
verified present after the rebase, and the rebase conflicts — the `state == "success"` predicate
and the `1./1.` list markers — were resolved by taking #1542's merged fix.

**4. Authorization.** The operator authorised merging the open PRs without a completed CodeRabbit
review today, because CodeRabbit was account-wide rate-limited. This PR is additionally released
from the lead review gate in its body: the lead read the rebase result and the budget measurement
above and cleared it.

PR #1562, opened 2026-08-15
docs(readme): say the companion sessions are launched by hand

GoosLab:
Follow-up to #1561, which merged while this correction was being applied.

CodeRabbit's review of #1561 flagged that the replacement callout reads as though the launcher starts all five sessions:

> The README may imply that Kanban launches all companion sessions automatically, although the launch switches start only the lead session and the companions require separate launches.

That is right, and it contradicts the v3.1 section two hundred lines earlier in the same file, which states plainly that companion sessions are launched by hand, one per terminal, and that a session never spawns a peer.

Reworded in all four locales: `moai cc -k` (or `moai glm -k`) starts the lead, and `-k --name <role>-<run-id>` joins each companion — launched by hand, one per terminal.

Verified against `moai cc --help` on the installed `3.1.0` binary, which documents the two forms separately: `-k, --kanban [SPEC-ID]` to enter as the lead, and `-k --name <role>-<run-id>` to join an existing run as a companion.

## Verification

- Four files, one line each; `git diff origin/main --stat` → 4 insertions, 4 deletions
- `grep -c '^## '` unchanged: en/ja/zh 14, ko 11 — parity-neutral
- docs-site untouched

🗿 MoAI

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -393,7 +393,7 @@ flowchart TD
 | CWD 衝突解決 | `(worktree_path, session_id)` の組が再利用パスを区別 |
 | 深さ上限（depth ceiling） | ネスト複雑度を制限 |
 
-> **現在利用可能**: `moai cc -k` / `moai glm -k` がリードと4つの同伴セッションを起動し、`moai chain <status|lineage|back|list|prune>` が系譜を読み、`moai todo <add|list|next|done>` が `backlog` カラムを操作します。起動手順は上の「v3.1 の新機能 — カンバンモード」節にあります。
+> **現在利用可能**: `moai cc -k`（または `moai glm -k`）がリードを起動し、`-k --name <role>-<run-id>` で同伴セッションを1つずつ参加させます — ターミナルごとに1つ、手で起動します。`moai chain <status|lineage|back|list|prune>` が系譜を読み、`moai todo <add|list|next|done>` が `backlog` カラムを操作します。起動手順は上の「v3.1 の新機能 — カンバンモード」節にあります。
 
 > 詳細: [カンバンモードガイド](https://adk.mo.ai.kr/ja/advanced/kanban-mode)
 
--- a/README.ko.md
+++ b/README.ko.md
@@ -230,7 +230,7 @@ SPEC마다 독립된 작업 트리를 준다. `moai cc -w <이름>`으로 진입
 | CWD 충돌 해결 | `(worktree_path, session_id)` 쌍으로 재사용 경로를 구분 |
 | 깊이 상한 | 중첩 복잡도를 제한 |
 
-> **지금 쓸 수 있습니다**: `moai cc -k` / `moai glm -k`로 리드와 네 개의 동반 세션을 띄우고, `moai chain <status|lineage|back|list|prune>`으로 계보를 읽고, `moai todo <add|list|next|done>`으로 `backlog` 컬럼을 운영합니다. 실행 순서는 위 "v3.1 새 기능 — 칸반 모드" 절에 있습니다.
+> **지금 쓸 수 있습니다**: `moai cc -k`(또는 `moai glm -k`)로 리드를 띄우고, `-k --name <role>-<run-id>`로 동반 세션을 하나씩 붙입니다 — 터미널당 하나씩 손으로 띄웁니다. `moai chain <status|lineage|back|list|prune>`으로 계보를 읽고, `moai todo <add|list|next|done>`으로 `backlog` 컬럼을 운영합니다. 실행 순서는 위 "v3.1 새 기능 — 칸반 모드" 절에 있습니다.
 
 > 자세히: [칸반 모드 가이드](https://adk.mo.ai.kr/ko/advanced/kanban-mode)
 
--- a/README.md
+++ b/README.md
@@ -397,7 +397,7 @@ flowchart TD
 | CWD-collision resolution | `(worktree_path, session_id)` pair disambiguates reused paths |
 | Depth ceiling | Caps nesting complexity |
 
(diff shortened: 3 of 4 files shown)


GoosLab:
## Merge-without-review evidence

Posted before merge, because CodeRabbit did not review this head.

### 1. Review status — NOT reviewed

```
$ gh api repos/modu-ai/moai-adk/commits/a83be9f8b/status \
    --jq '.statuses[] | select(.context == "CodeRabbit") | "\(.state) :: \(.description)"'
success :: Review rate limited
```

The `gh pr checks` row reads `CodeRabbit pass`, which is the false-green case: the status state is `success` on an unreviewed head exactly as it is on a reviewed one. Only the description separates them, and here it reads `Review rate limited` — the review never started. There is no `Merge Risk:` line for `a83be9f8b`.

For contrast, the parent PR #1561 *was* reviewed: `success :: Review completed`, with `Merge Risk: 🔵 Low · up to 396a3` matching that head. This PR exists because of a finding from that review.

### 2. CI — green, with a stated limit

`gh pr checks 1562 --watch --fail-fast` exited 0. Builds passed on all five targets (darwin amd64/arm64, linux amd64/arm64, windows amd64); `Lint` 2m36s, `Constitution Check` 1m11s, `Analyze (Go)` pass.

**This is not suite evidence.** `Test (ubuntu-latest)` passed in **3 seconds** because `Detect (paths-filter)` skipped the Go jobs on a docs-only diff — `Race Test`, `Integration Tests`, and the `Test (matrix.os)` legs are all `skipping`. Reading this CI green as "the test suite passed" would repeat the #1547 gap.

### 3. What was verified by hand

```
$ git diff origin/main --numstat
1  1  README.ja.md
1  1  README.ko.md
1  1  README.md
1  1  README.zh.md

$ grep -c '^## ' README.md README.ko.md README.ja.md README.zh.md
README.md:14   README.ko.md:11   README.ja.md:14   README.zh.md:14
```

Identical to `origin/main` (ko's 11-vs-14 divergence is pre-existing and out of this card's scope), so the change is locale-parity-neutral. All four locales move together in this PR, and one line changes in each. docs-site is untouched, so the hugo build gate is unaffected.

### 4. Full card result table

The card claimed "8 README defects confirmed" and named one. The other seven were not recoverable — `.moai/reports/backlog/` (10 files), `.moai/reports/`, the kanban backlog queue (24 cards), and the last 15 merged PRs carry no such list. Rather than invent seven, the four files were audited directly. The numbering below is that audit's, not the card's.

| # | Finding | Verdict | Evidence |
|---|---|---|---|
| 1 | Kanban "later phases" self-contradiction — the reference section called the `-k` switch, the `moai chain` CLI, and the multi-session board columns "planned for later phases" while the same file documents their launch sequence 90 lines earlier | **Fixed** (#1561, 4 locales) | Against the installed `3.1.0` (`ed04e40e6`): `moai cc --help` → `-k, --kanban [SPEC-ID]` + `-k --name <role>-<run-id>`; `moai chain --help` → `status\|lineage\|back\|list\|prune`; `moai todo --help` → `add\|list\|done\|next`. The linked `advanced/kanban-mode.md` had already been rewritten, leaving the README as the last carrier |
| 2 | Slash-command count said 15 and omitted `todo` | **Fixed** (#1561, 4 locales) | 16 actual. The same README uses `/moai todo` in its v3.1 section and links its docs page |
| 3 | "Full 36 commands" | **Fixed** (#1561, en/ja/zh; ko lacks the line) | `moai --help` lists 47 entries, 45 excluding the `help` and `completion` cobra built-ins. The count was dropped rather than replaced, since a fresh number goes stale the same way |
| 4 | Stale MCP paragraph carried only by ja and zh — claimed the distributed `.mcp.json` bundles three third-party servers by default | **Fixed** (#1561, ja/zh deleted) | English (canonical) had already dropped it. `internal/template/templates/.mcp.json` contains exactly one entry (`moai`), which is what the MCP section of those same files says |
| 5 | `moai mcp add\|remove\|list` documented in all four locales but unreachable | **Confirmed — code defect, not a docs defect** | `newMCPCmd()` at `internal/cli/mcp.go:60`, referenced only by `mcp_test.go:427`; never registered on root. `moai mcp` → `Unknown command`. Left untouched deliberately: rewriting the README to describe hand-editing `.mcp.json` would freeze the bug as the specification. Raised as a separate card, and the real problem is that the test calls `newMCPCmd()` directly, so it cannot catch non-registration |
| 6 | `README.ko.md` is a structural fork — 11 H2 against 14, different section set | **Rejected — out of scope** | `origin/main` measures the same 11 vs 14, so it is pre-existing. Backlogged with #7 |
| 7 | ja's Cost section diverges from en/zh — extra intro H3, no `DeepSWE Benchmark` heading, Verification Economy / Budget Defense split in two | **Confirmed — not fixed** | Same class of structural debt as #6; backlogged together |
| 8 | `kanban_chain` reads `plan → run → verify → sync` while the board columns are `plan → run → review → sync` | **Rejected — premise sits outside the README** | `moai cc --help` carries both spellings itself (chain says `verify`, roles say `review`). The README is not wrong relative to the code; the inconsistency is upstream in the CLI |

Also checked and judged clean: URL blacklist (zero hits across the four files); the 🪙🧠🛡️ heading emoji (the no-emoji rule targets docs-site, where the icon shortcode exists — README is not Hugo); the statusline example's `🗿 v3.0.1` (a real version pair, an illustration rather than a claim).

### 5. This PR's own change

CodeRabbit's review of #1561 flagged that the replacement callout reads as though the launcher starts all five sessions. It does not — it starts the lead, and each companion joins separately with `-k --name <role>-<run-id>`, one per terminal, which is what the v3.1 section states 200 lines earlier. Reworded in all four locales.

### 6. Authorization

Merge approved by the kanban lead on the basis that the change is docs-only, one line per file, locale-parity-verified, and that the review absence is recorded here rather than papered over. This comment does not substitute for the review — it records the judgement made in its absence.

🗿 MoAI

PR #1564, opened 2026-08-15
ci(release): drop the duplicate ubuntu race run from release PR verification

GoosLab:
## What this removes

`release-range-verify` is deleted, and `release-pr-gate` no longer waits on it.

## Why — the job did not do what its name said

The job's own comment claimed it verified "the accumulated release range (since the last tag) ... NOT just the tip commit". Its steps were:

| Step | What it actually did |
|---|---|
| `Checkout code` | `actions/checkout@v7`, `fetch-depth: 0`, **no `ref:` override** — on a `pull_request` event this is the PR merge ref, the same tree every other job checks out |
| `Setup Go` / `Install ast-grep` / `Verify sg` | identical to the `full-matrix-test` ubuntu leg |
| `Print commit range since last tag` | `git describe --tags --abbrev=0` then `git log <tag>..HEAD --oneline` — **prints only** |
| `Run go build (release range)` | `go build ./...` at HEAD |
| `Run go test with race detector (release range)` | `go test -race -timeout 25m ./...` at HEAD |

No step ever checked out an intermediate commit. The "range" existed only in the `git log` output. The verification it performed was the ubuntu leg of `full-matrix-test`, byte for byte, with a `git log` and a `go build ./...` in front of it — and `go build` is already covered by ci.yml's cross-compilation `Build` matrix and by `go vet ./...`, which is what compiles `_test.go` per target.

## Measurement

v3.1.0 release PR #1555, successful runs `31888268855` (Multi-OS) and `31888268858` (CI):

```
Multi-OS │ Release Verify (windows-latest)   796s
         │ Release Verify (ubuntu-latest)    550s   ← ubuntu race
         │ Release Range Verify              549s   ← ubuntu race, removed here
         │ Release Verify (macos-latest)     506s
         │ Detect / Gate                     3s / 4s
CI       │ Race Test                         565s   ← ubuntu race, advisory
         │ Test (ubuntu-latest)              300s
         │ Lint 155s · Build ×5 76–89s · Integration ×3 43–76s
```

The same `go test -race ./...` ran on ubuntu three times: **1,664 runner-seconds, of which 549 bought nothing.**

The duplication only fires on `release/*` PRs — `detect-release` gates the workflow with `if: startsWith(github.head_ref, 'release/')`, so ordinary PRs skip it in 6–8s.

## Honest limitation — this is cost, not latency

Wall-clock is set by the windows leg at 796s, so the workflow still takes **~800s instead of 815s**. What this buys is runner minutes and queue occupancy, not a shorter wait for a release. PR #1555 ran this workflow six times (790s–1365s each), so the saving is real in aggregate — but anyone hoping this shortens release-day feedback should know the actual target is the **windows leg at 796s**, which this PR does not touch.

## Why `ubuntu-latest` stays in the matrix

An earlier draft of this change also trimmed the matrix to `[macos-latest, windows-latest]`, on the strength of the job comment at line 43 — *"macOS + Windows (ubuntu already covered by main CI)"*. Checking branch protection first showed that would have been wrong:

```
$ gh api repos/modu-ai/moai-adk/branches/main/protection --jq '.required_status_checks.contexts'
["Test (ubuntu-latest)","Lint","Build (linux/amd64)","Analyze (Go) (go)","Release PR Multi-OS Gate"]
```

ci.yml's `Race Test` **is** the ubuntu race coverage on every PR, but its check name is deliberately absent from that list — its own comment says so: *"non-blocking because 'Race Test' is a NEW check name NOT in branch-protection required checks"*. It is visible and advisory; it cannot stop a merge.

`Release PR Multi-OS Gate` **is** required, and it reports this workflow. That makes the ubuntu leg here the only ubuntu race run that can block a release. Trimming it would have quietly demoted ubuntu race coverage to advisory at exactly the moment it matters most.

So the code was right and the comment was wrong. The comment is corrected to state the real reason ubuntu is in the matrix, rather than the matrix being trimmed to match a stale comment.

## Secondary — a latent SIGPIPE defect in the same file

Found by the lane fixing the release provenance gate (#1563), where the same construct was failing in production rather than lying dormant.

`shell: bash` runs a step under `--noprofile --norc -eo pipefail`. `grep -q` exits at its first match and closes the pipe, so a producer with more to write dies of SIGPIPE, and under `pipefail` the pipeline resolves to **141** even though the match succeeded. Reproduced locally on this tree:

```
$ bash --noprofile --norc -eo pipefail -c 'git log --oneline | grep -qi "readme"'
exit 141
$ bash --noprofile --norc -eo pipefail -c 'git log --oneline | grep -i "readme" >/dev/null'
exit 0
```

Two call sites existed in this file:

- **line 92**, in the surviving `full-matrix-test` job — **fixed here** (`grep -qi` → `grep -i ... >/dev/null`, which keeps grep reading to EOF; exit semantics are unchanged).
- **line 148**, in `release-range-verify` — **resolved by the job deletion**, no separate fix needed.

`sg --version` prints a single short line, so the producer finishes inside the pipe buffer and neither site trips today. The form is corrected because that guarantee rests on output length rather than on anything the check controls.

`ci.yml:121` carries the same construct and is **not** touched here — it is assigned to the lane handling #1563.

## Follow-up observations (not in this PR)

- `release-pr-multi-os.yml` has no paths-filter, so a docs-only `release/*` PR still runs the full ~13-minute matrix. Low priority given how rarely release PRs are docs-only.
- ast-grep is `npm install -g`'d six times per release PR (3 OS legs + range-verify + ci.yml `test` + ci.yml `test-race`); ~10–20s each. After this PR, five.
- On a release PR, ci.yml's advisory `Race Test` (565s) still duplicates the blocking ubuntu leg. Gating it with `if: !startsWith(github.head_ref, 'release/')` would save another ~565s with no loss of gate strength — but that edit lands in `ci.yml`, outside this PR's file scope.

## Verification

```
$ python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/release-pr-multi-os.yml')); ..."
jobs: ['detect-release', 'full-matrix-test', 'release-pr-gate']
matrix: ['ubuntu-latest', 'macos-latest', 'windows-latest']
gate needs: ['full-matrix-test']

$ grep -rn "release-range-verify\|Release Range Verify" .github/
(no matches — no dangling reference)
```

The required check name `Release PR Multi-OS Gate` is unchanged, so branch protection needs no edit.

Scope: one file, +25 / −100.

🗿 MoAI

### DIFF
--- a/.github/workflows/release-pr-multi-os.yml
+++ b/.github/workflows/release-pr-multi-os.yml
@@ -40,8 +40,17 @@ jobs:
         run: echo "proceed=true" >> $GITHUB_OUTPUT
 
   # ---------------------------------------------------------------------------
-  # Full 3-OS matrix test: macOS + Windows (ubuntu already covered by main CI)
-  # Only runs for release/* branch PRs or manual workflow_dispatch
+  # Full 3-OS matrix test. Only runs for release/* branch PRs or manual
+  # workflow_dispatch.
+  #
+  # ubuntu-latest stays in the matrix even though ci.yml also runs the race
+  # detector on ubuntu. ci.yml's `Race Test` job is advisory by construction —
+  # its check name is deliberately absent from branch protection, so a race
+  # failure there is visible but does not block a merge. This job reports
+  # through `Release PR Multi-OS Gate`, which IS a required check, so the
+  # ubuntu leg here is the only ubuntu race run that can stop a release.
+  # Dropping it would silently demote ubuntu race coverage to advisory at
+  # exactly the moment it matters most.
   # ---------------------------------------------------------------------------
   full-matrix-test:
     name: Release Verify (${{ matrix.os }})
@@ -86,10 +95,18 @@ jobs:
       # Ubuntu and Debian ship /usr/bin/sg as a newgrp alternative, so `sg`
       # being on PATH proves nothing about what it is. Assert the binary is
       # ast-grep — a silent substitution is the failure this step prevents.
+      #
+      # `grep -i ... >/dev/null` rather than `grep -qi`: `shell: bash` runs the
+      # step under `-eo pipefail`, and `grep -q` closes the pipe at the first
+      # match, so a producer with more to write dies of SIGPIPE and the
+      # pipeline resolves to 141 even though the match succeeded. Redirecting
+      # instead keeps grep reading to EOF, so the exit status means what it
+      # reads as. `sg --version` prints one short line and so does not trip
+      # this today; the form is fixed because the next producer might.
       - name: Verify `sg` resolves to ast-grep
         shell: bash
         run: |
-          if ! sg --version 2>&1 | grep -qi 'ast-grep'; then
+          if ! sg --version 2>&1 | grep -i 'ast-grep' >/dev/null; then
             echo "FAIL: 'sg' on PATH is not ast-grep — the rewrite guards cannot run."
             echo "resolved to: $(command -v sg || echo '<not found>')"
             sg --version 2>&1 || true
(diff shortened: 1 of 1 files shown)


GoosLab:
### Review evidence (automated review unavailable on this head)

CodeRabbit reports `success` on `cdfb2ef9603e914a3b99ca0c41cb3b0b418cb482`, but its status description is `Review rate limited` — not `Review completed`. The status state is identical in both cases, so the green row is **not** evidence that a review ran. Recording the manual verification in its place.

```
$ gh api repos/modu-ai/moai-adk/commits/cdfb2ef9603e914a3b99ca0c41cb3b0b418cb482/status \
    --jq '.statuses[] | select(.context == "CodeRabbit") | "\(.state) | \(.description)"'
success | Review rate limited
```

**Structural checks on the edited file**

```
$ python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/release-pr-multi-os.yml')); ..."
jobs:        ['detect-release', 'full-matrix-test', 'release-pr-gate']
matrix:      ['ubuntu-latest', 'macos-latest', 'windows-latest']
gate needs:  ['full-matrix-test']

$ grep -rn "release-range-verify\|Release Range Verify" .github/
(no matches)
```

No dangling reference to the deleted job, and `Release PR Multi-OS Gate` — the required check name — is unchanged, so branch protection needs no edit.

**Gate behaviour observed on this PR**

`Release PR Multi-OS Gate` reported `pass` in 3s via the non-release skip path (`Detect Release PR` → `skipping`, `Release Verify (${{ matrix.os }})` → `skipping`), which is the correct behaviour for a non-`release/*` PR and confirms the gate still reports after losing one of its two `needs`.

**What is NOT verified here**

This PR's own CI cannot exercise the changed path: `detect-release` gates on `startsWith(github.head_ref, 'release/')`, and this branch is `worktree-card-ci-dedup`. The first real exercise of the trimmed matrix will be the next `release/*` PR. The residual risk is bounded — the deletion removes a job and rewires one `needs` list; it adds no new logic — but it is a genuine gap rather than something these green checks cover.

Go tests did not run on this PR either: `.github/workflows/release-pr-multi-os.yml` is absent from ci.yml's `go_code` paths-filter, so `Detect (paths-filter)` resolved false and `Test (ubuntu-latest)` passed in 3s via the skip-marker job. That is correct for a workflow-only change, but "CI green" here does not mean the Go suite ran.

🗿 MoAI

PR #1569, opened 2026-08-16
fix(profile): match the launch ledger by directory, not by path spelling

GoosLab:
## Problem

Entering a project through a differently-cased path silently sent the launch to the **default** profile, and with it to a different transcript store — so `moai cc -c` / `-r` found no prior session and the project's history looked lost.

`launch.yaml` records a project under the spelling the launcher saw:

```yaml
projects:
    /Users/goos/MoAI/moai-adk-go: moai-adk
```

On a case-insensitive filesystem `cd .../moai/moai-adk-go` and `cd .../MoAI/moai-adk-go` open the same directory, but `os.Getwd` returns whichever spelling the shell used and `filepath.EvalSymlinks` does **not** canonicalize case. Two spellings, two keys, lookup miss.

## Fix

`lookupProjectKey` falls back to `os.SameFile` when the exact string misses — the filesystem decides whether two spellings name one directory, instead of this code guessing a per-OS case rule. Case-sensitive filesystems keep working: there `moai/` and `MoAI/` really are two directories and `SameFile` says no.

The write side reuses an already-recorded spelling, so one directory never accumulates two entries that later drift apart.

## Verification

A/B against the pre-fix binary:

```
=== PWD=/Users/goos/moai/moai-adk-go  (lowercase) ===
  old: default        <- the split
  new: moai-adk
=== PWD=/Users/goos/MoAI/moai-adk-go  (canonical) ===
  old: moai-adk
  new: moai-adk
```

`go test ./internal/profile/ ./internal/cli/` — ok. Four new tests:

| Test | Guards |
|---|---|
| `CaseVariantPath` | case-variant path resolves to the same profile |
| `CaseVariantReusesKey` | one directory keeps one ledger entry |
| `DistinctDirectoriesDoNotMatch` | a genuinely different directory never matches |
| `MissingDirectoryDoesNotMatch` | a deleted path matches nothing |

The two case-fold tests skip on a case-sensitive filesystem rather than assert behavior the platform cannot exhibit.

🗿 MoAI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project profile recognition for paths that differ only by letter casing.
  * Prevented duplicate project entries when equivalent path spellings are used.
  * Ensured distinct or missing directories are not incorrectly treated as the same project.
  * Made matching deterministic when multiple equivalent entries exist.
  * Added warnings for conflicting profiles and safely handled invalid project data.

* **Tests**
  * Added coverage for case variants, duplicate entries, conflicting profiles, distinct directories, and missing directories.

### DIFF
--- a/internal/profile/profile.go
+++ b/internal/profile/profile.go
@@ -275,6 +275,48 @@ func normalizeProjectKey(projectRoot string) string {
 	return cleaned
 }
 
+// lookupProjectKey finds the ledger key that names the SAME directory as key,
+// returning the stored key and whether one was found.
+//
+// An exact string match wins. When it misses, the map is scanned and each
+// candidate is compared with os.SameFile — the filesystem itself decides
+// whether two spellings name one directory, rather than this code guessing a
+// per-OS case rule.
+//
+// The case-variant miss is the failure this exists for. On macOS and Windows
+// the filesystem is case-insensitive, so `cd /Users/x/moai/repo` and
+// `cd /Users/x/MoAI/repo` open the same directory, but os.Getwd reports back
+// whichever spelling the shell used and filepath.EvalSymlinks does NOT
+// canonicalize case. The two spellings therefore hash to different ledger keys,
+// the lookup misses, and the launch silently falls back to the default profile
+// — which stores its transcripts elsewhere, so `--continue` / `--resume` find
+// no prior session. os.SameFile keeps the lookup correct on case-sensitive
+// filesystems too, where two same-spelled-but-differently-cased directories can
+// genuinely both exist and must NOT be conflated.
+func lookupProjectKey(projects map[string]any, key string) (string, bool) {
+	if key == "" || projects == nil {
+		return "", false
+	}
+	if _, ok := projects[key]; ok {
+		return key, true
+	}
+
+	info, err := os.Stat(key)
+	if err != nil {
+		return "", false
+	}
+	for candidate := range projects {
+		candidateInfo, err := os.Stat(candidate)
+		if err != nil {
+			continue
+		}
+		if os.SameFile(info, candidateInfo) {
+			return candidate, true
+		}
+	}
+	return "", false
+}
+
 // loadLaunchLedger reads and decodes the ledger as a generic map so unknown and
 // legacy keys are tolerated rather than rejected. A missing or corrupt file
 // yields an error the callers translate into "no fallback".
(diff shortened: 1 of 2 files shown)


GoosLab:
Addressed the Merge Risk verdict. One of the two concerns was real and is now fixed; the other does not arise on the launch path.

**1. Duplicate entries chosen unpredictably — real, fixed in 92bcaf9b7.**

The concern is exactly right, and it is not hypothetical: a ledger written *before* this fix can already hold two spellings of one directory, because the case-spelling miss is what produced them. Go randomizes map iteration, so the unordered scan handed back either entry from run to run — the same project resolving to a different profile, and therefore a different transcript store, on each launch.

The scan now walks the candidate keys sorted, so the winner is stable, and the write side collapses the duplicate to one entry on the next launch.

The new test is falsifiable rather than decorative: with the `sort.Strings` call neutralized it trips at iterations 9 and 13 across five runs; with it restored, 50 iterations × 5 runs are stable.

**2. Ledger entry whose directory no longer exists — cannot arise here.**

The lookup key is not arbitrary input: it is `normalizeProjectKey(findProjectRoot())`, the root of the checkout the launcher is *currently running in*, which was reached by stat-ing `.moai` on the way up. A key that does not resolve means the process has no working directory.

The scan path stats the lookup key anyway and returns no match when it fails, and candidate keys that do not stat are skipped. Beyond that, the resolved profile is validated by the pre-existing `launchCandidateIsUsable` before it is returned, so a stale *profile* directory is already refused. Adding a stat to the exact-match branch would guard a state the launch path cannot reach, so it is left as-is rather than carried as unreachable code.

🗿 MoAI

GoosLab:
Second pass on the Merge Risk verdict. One point led to a further change; the other I am closing with evidence rather than code.

**Conflicting equivalent keys — addressed in fb5f6bbeb.**

The point stands and the previous commit did not fully answer it: sorting fixed *which* duplicate wins, but stability is not correctness. Where two equivalent keys name different profiles, the winner is stable and still possibly not the one the user meant — the ledger records no write order, so nothing in the data says which entry is current.

Ledger cleanup cannot be inferred, because "the right entry" is not derivable from what is stored. So the ambiguity is now surfaced once on stderr, naming both profiles and the ledger path, rather than routing the user somewhere silently:

```
Warning: launch.yaml records "/path/to/project" twice for one directory, under
different profiles ("moai-adk" vs "other"). Using "moai-adk". Remove the stale
entry from ~/.moai/claude-profiles/launch.yaml to silence this.
```

Duplicates that agree on one profile are unambiguous and stay silent — asserted by a sibling subtest, so the warning cannot degrade into noise. Ledger values are also read through a string assertion yielding `""` rather than panicking, so a hand-edited entry cannot crash resolution.

**Deleted project paths — measured, not reachable on this path.**

The lookup key is `normalizeProjectKey(findProjectRoot())`: the root of the checkout the process is *currently running in*, reached by stat-ing `.moai` on the way up. A key that does not resolve means the process has no working directory.

Downstream, the returned profile name is validated by the pre-existing `launchCandidateIsUsable` before resolution returns it, so a stale *profile* directory is already refused — that is the failure that would actually cost a user their prior sessions, and it is guarded. `TestLookupProjectKey_MissingDirectoryDoesNotMatch` covers the scan side: a nonexistent lookup misses, and a ghost candidate key is skipped rather than matched.

Adding a stat to the exact-match branch would guard a state this path cannot reach, so it stays out rather than shipping as unreachable code.

🗿 MoAI

PR #1566, opened 2026-08-15
fix(cli): register moai mcp on root, and guard the registration

GoosLab:
Found while auditing the four READMEs in #1561: all of them tell users to enable the disabled MCP entries with `moai mcp add <name>`, and add that users "never hand-edit `.mcp.json`". Neither was true — `moai mcp` did not exist as a runnable command.

```
$ moai mcp
  ERROR  Unknown command "mcp" for "moai"

$ grep -rn 'newMCPCmd' --include='*.go' .
  internal/cli/mcp.go:60        (definition)
  internal/cli/mcp_test.go:427  (its own test)
  → zero registrations on root
```

`newMCPCmd()` builds a complete `add` / `remove` / `list` tree over the same atomic-RMW seam the GLM tools CLI uses, and fourteen tests cover its behaviour. Nothing added it to `rootCmd`. The command was written, tested, documented, and unreachable.

## Why the tests did not catch it

`TestNewMCPCmd_Registered` — the name promises registration coverage — called `newMCPCmd()` directly and asserted the factory's own children. It never asked *who calls the factory*, so it passed identically whether or not root registered anything. That is the actual defect worth fixing: a one-line omission stayed invisible because the test that would have caught it was testing the wrong side of the boundary.

Two changes:

- Renamed it to `TestNewMCPCmd_FactoryChildren`, so its name matches what it checks. It still earns its place — factory wiring and root wiring are different questions.
- Added `TestMCPCmdRegisteredOnRoot`, which walks the real `rootCmd` tree, finds `mcp`, and asserts its three verbs.

## Verified the guard fails for the right reason

With the `AddCommand` line removed:

```
--- PASS: TestNewMCPCmd_FactoryChildren (0.00s)
--- FAIL: TestMCPCmdRegisteredOnRoot (0.00s)
    mcp_test.go:480: `moai mcp` is not registered on rootCmd — README and docs-site
    document `moai mcp add <name>` as the way to enable MCP entries, so an
    unregistered factory makes a documented surface unreachable.
    Add newMCPCmd() in root.go next to newMCPServerCmd(). Registered: [agent
    ast-edit ast-grep cc cg chain clean config constitution doctor epic gate
    github glm goal handoff harness hook init inventory loop lsp mcp-server ...]
```

The old test passing beside the new one failing is the gap itself, reproduced. Line restored afterwards; the diff below is the restored state.

## All three verbs work end to end

Built from this branch, run against a scratch project root so no tracked `.mcp.json` was touched:

```
[...]

PR #1580, opened 2026-08-17
fix(config): skip config-cache write in uninitialized directories so init validation passes

### LINKED ISSUE #1568: [Bug] moai init fails with "project already initialized" in a brand-new empty directory (v3.1.0 regression)
## Describe the Bug

`moai init` fails in a brand-new, empty directory, aborting with **"project already initialized"** — so a fresh project cannot be initialized at all in v3.1.0.

The init run itself writes a `.moai/` cache into the directory during startup/detection, and then the later "Validating project structure" step detects that `.moai/` and incorrectly concludes the project is already initialized. `moai init --force` works, and `moai update` on existing projects is unaffected — the problem is specific to a first-time `moai init`.

## To Reproduce

1. Create and enter a brand-new empty directory:
   ```
   mkdir new-project && cd new-project
   ```
2. Run:
   ```
   moai init
   ```
3. See the run fail at the validation step with `project already initialized`.

## Expected Behavior

`moai init` initializes a new MoAI project in the empty directory (the behavior in v3.0.1).

## Actual Behavior

Init aborts at "Validating project structure" with "project already initialized", even though the directory started empty. A partial `.moai/` — specifically `.moai/state/config-cache.json` — is created earlier in the same run and left behind; that self-written `.moai/` appears to be what the validation step then flags as a pre-existing project.

## Environment

- **OS**: Ubuntu 24.04 (Linux x86_64)
- **Go Version**: N/A (official prebuilt release binary)
- **MoAI-ADK Version**: 3.1.0
- **Claude Code Version**: N/A (reproduced directly with the `moai` CLI)
- **Shell**: bash

## Logs / Error Output

```
· Initializing MoAI project...
  ● Detection: Analyzing project structure...
  ✓ Detected project structure
  ● Validation: Using mode: tdd...
  ✓ Mode validated
  ● Validation: Validating project structure...
  ✗ project already initialized

   ERROR

  Initialization failed: project already initialized
    Hint: this directory already contains a MoAI project — did you mean 'moai update' (refresh templates in place)?
  Re-run with --force only to reinitialize from scratch.
```

After the failure, the "empty" directory contains only:

```
new-project/.moai/state/config-cache.json
```

## Additional Context

- **Regression**: v3.0.1 does **not** exhibit this — `moai init` in a fresh empty directory succeeds. The init path was restructured for 3.1.0 (single-wizard rewrite), which is the likely origin of the ordering change.
- **Workaround**: `moai init --force` completes successfully in the same empty directory (nothing to back up, so it produces a clean project).
- **Unaffected**: `moai update` on already-initialized projects works normally.
- **Likely root cause**: the config-cache write (`.moai/state/config-cache.json`) happens *before* the "already initialized" validation, so validation detects the tool's own freshly-written `.moai/`. Deferring that write until after validation — or excluding it from the "already initialized" check — should fix it.

GoosLab:
### Summary

`moai init` in a brand-new empty directory failed with "project already
initialized" (issue #1568, a v3.1.0 regression). The cause is an ordering
side effect: every non-trivial command eagerly loads the project config
(`InitDependencies` → `ConfigManager.Load` → `Loader.LoadWithCache`), and on
a cache miss `writeCache` ran `os.MkdirAll(<root>/.moai/state/)`,
materializing `.moai/` in the uninitialized directory before init's
validator ran. The validator then flagged the tool's own artifact as an
existing project and aborted.

Two changes:

1. **Root cause** — `internal/config/cache.go`: `LoadWithCache` skips the
   cache write when the config directory itself does not exist. An
   uninitialized directory is left untouched (this also stops every other
   non-trivial moai command from leaving `.moai/state/config-cache.json`
   litter in arbitrary working directories). Caching resumes unchanged once
   the directory exists (after init/update).
2. **Poison healing** — `internal/core/project/validator.go`: a `.moai/`
   containing ONLY `state/config-cache.json` is treated as tool litter, not
   an initialized project, so directories already poisoned by a pre-fix
   binary initialize cleanly after upgrading — no manual cleanup, no
   `--force`. Any other content in `.moai/` still fails validation exactly
   as before.

### Root-cause archaeology

The cache layer landed with the config-perf work in the v3.1.0 cycle; the
eager `deps.Config.Load(cwd)` startup wiring is what puts it ahead of init's
`Validate` call. v3.0.1 had no disk-cache layer, hence no regression there.

### Test plan

- New regression test `TestCache_DoesNotCreateConfigDir`
  (`internal/config/cache_test.go`): asserts `LoadWithCache` on a
  nonexistent config dir creates nothing. Demonstrated FAILING on
  origin/main (fix reverted, test kept) and PASSING with the fix.
- New validator table cases (`internal/core/project/validator_test.go`):
  cache-artifact-only `.moai/` → valid (the poisoning case, fails on
  origin/main); artifact + any real content → still invalid; unexpected
  state content → still invalid (controls).
- Binary-level, both directions: origin/main build in a fresh empty dir →
  exit 1 "project already initialized" with `.moai/state/config-cache.json`
  (8,197 bytes) left behind (the reporter's exact symptom); fixed build in
  the same scenario → exit 0, project initialized. Fixed build also
  succeeds in a directory pre-poisoned with the artifact.
- `go test ./internal/config/ -run TestCache -count=1` → ok;
  `go test ./internal/core/project/ -count=1` → ok. The only failures in
  the full package runs (`TestShippedConfigKeysHaveReaders` in config) are
  pre-existing on pristine origin/main, verified on an untouched extraction.

Fixes #1568

### DIFF
--- a/internal/config/cache.go
+++ b/internal/config/cache.go
@@ -122,6 +122,19 @@ func (l *Loader) LoadWithCache(configDir string) (*Config, error) {
 		return nil, err
 	}
 
+	// Skip the cache write when the config directory itself does not exist
+	// (issue #1568): writeCache's MkdirAll would materialize
+	// <root>/.moai/state/ as a side effect, and `moai init` then fails its
+	// "project already initialized" validation against the directory this
+	// very command just created. The project is not initialized here, so
+	// there is nothing worth caching; caching resumes once the directory
+	// exists (after init or update).
+	if _, statErr := os.Stat(configDir); statErr != nil {
+		slog.Debug("config cache: skipping write (config directory does not exist)",
+			"config_dir", configDir)
+		return cfg, nil
+	}
+
 	// Write cache (fail-open: a write failure is logged and ignored — the
 	// cache is an optimization, not a correctness dependency; C-EDGE-003).
 	if writeErr := writeCache(cachePath, cfg, currentFP); writeErr != nil {
--- a/internal/config/cache_test.go
+++ b/internal/config/cache_test.go
@@ -297,6 +297,33 @@ func TestCache_LocationUnderStateDir(t *testing.T) {
 	}
 }
 
+// TestCache_DoesNotCreateConfigDir (issue #1568) verifies that a cache-miss
+// write does not materialize the config directory as a side effect. Before
+// this guard, running any non-trivial moai command in a fresh directory
+// created <dir>/.moai/state/config-cache.json via writeCache's MkdirAll, and
+// `moai init` then failed its "project already initialized" validation
+// against the directory the very same command had just created.
+func TestCache_DoesNotCreateConfigDir(t *testing.T) {
+	requireCacheEnabled(t)
+	base := t.TempDir()
+	// The config directory of an uninitialized project: <root>/.moai, absent.
+	configDir := filepath.Join(base, "fresh-project", ".moai")
+
+	cfg, err := NewLoader().LoadWithCache(configDir)
+	if err != nil {
+		t.Fatalf("LoadWithCache on uninitialized dir: %v", err)
+	}
+	if cfg == nil {
+		t.Fatal("LoadWithCache returned nil config")
+	}
+
+	if _, statErr := os.Stat(configDir); statErr == nil {
+		t.Fatalf("LoadWithCache created %s as a side effect; the cache write must be skipped when the config directory does not exist", configDir)
+	} else if !os.IsNotExist(statErr) {
+		t.Fatalf("stat %s: %v", configDir, statErr)
+	}
+}
+
 // TestCache_SizeChangeInvalidates (AP-2 defense) verifies that a file whose
 // mtime did NOT change but whose size DID is treated as invalid.
 func TestCache_SizeChangeInvalidates(t *testing.T) {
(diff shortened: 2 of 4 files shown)


GoosLab:
### fold 근거 (release/v3.1.1 통합 전 증거 코멘트 — #1549/#1566 선례 형식)

- **허브 리뷰 PASS (재측)**: 판정문 `.claude/worktrees/release-v311/.moai/reports/pr1580-pr1581-review-verdict.md` — 허브가 증거 self-report를 재측: 회귀테스트 `TestCache*`·`./internal/core/project/` 풀 ok (직접 실행), **바이너리 e2e 직접 재현** (fix 바이너리 빌드 → 빈 디렉터리 init → `INIT_EXIT=0` + `.moai/state/config-cache.json` 부재 확인). 4-관점 전부 PASS — 관용 범위는 `state/config-cache.json` 단일 shape로 협소(대조군 테이블 케이스 확인).
- **CI 전량 초록 (REST 확인)**: head `3f0077d6` check-runs 25/25 completed, failure 0 — 실패 0 실측.
- **CodeRabbit**: commit status `state=success` / `description=Review skipped` 실측 — 리뷰 미실시. 허브 판정문의 시한부 조항(#1566 선례) 3조건 — ① 허브 재측 PASS(판정문) ② CI 전량 초록(REST 확인) ③ 본 근거 코멘트 게시 — 로 fold 승인. CodeRabbit이 응답하면 원래 게이트 적용.

이 계기로 `release/v3.1.1`에 `merge --no-ff`로 통합합니다.

PR #1586, opened 2026-08-20
docs: reflect v3.1.1 across README 4-locale, CHANGELOG, and docs-site

GoosLab:
## What this is

v3.1.1's code landed in #1585 but the documentation had not caught up. This closes that, across four surfaces.

## The four commits

**`docs(changelog)`** — `[3.1.1]` carried a Summary and three SPEC-close bullets while ~30 user-facing changes went unrecorded, with no `Changed` or `Fixed` subsection at all. `release-drafter.yml` documents the procedure as "fold the drafted notes into CHANGELOG.md, then run `scripts/release.sh`", and `release.sh` lifts this section into the annotated tag message — so the gap would have become the release's permanent record. Adds 14 / 6 / 7 bullets; the Summary and the three SPEC bullets are untouched.

**`docs(readme)`** — a "what v3.1.1 adds" subsection (home hygiene, cross-session messaging config, prompt-cache config, `statusline.forge`, bare `/loop` as kanban foreman), plus the CLI table and config section extended, in all four locales. Authored in `README.ko.md` (canonical since 2026-08-17) and derived to en / ja / zh in the same change.

**`docs(docs-site)`** — the gaps left after `024536e38`, plus three menu omissions.

**`docs(reports)`** — the investigation behind it.

## Five counts were wrong

Not new drift — pre-existing, and measured rather than asserted this time:

| Claim | Was | Is | How measured |
|---|---|---|---|
| ref skills | 10 | **11** | `ls templates/.claude/skills/moai-ref-*` |
| domain skills | 5 in one place, 8 in another | **7** | same, both places were wrong |
| total CLI commands | 36 | **49** | `moai --help` top level, from a binary rebuilt off this tree |
| settings tabs (image alt) | 10 | **11** | body already named 11 |
| `.moai/config/sections/` | read as if 6 were all | **33 ship**, 6 commonly edited | directory count |

`MCP 21 tools`, the 12-agent catalog, GLM-5.3/1M, `manager-lead`, and the five-column board were checked and are correct — left alone. The 12-vs-11 agent tension is not an error either (11 custom + built-in `Explore`, which inherits the session model and has no profile cell); the README now says so.

## Two premises in the plan were wrong

Recorded rather than quietly fixed:

- **`cli-reference/{memory,tokens}` already existed** in all four locales with the content the plan said to write. The docs lane verified before creating and declined — creating them would have duplicated pages and broken the 160-file parity.
- **docs-site was further along than assumed.** `024536e38` had already landed 13 pages × 4 locales for v3.1.1. The plan's change inventory was built from code and CHANGELOG without consulting `git log -- docs-site/`.

One stale claim was corrected beyond the brief: `statusline.md` described forge counts as a fourth-line `2 / 1` segment, but `internal/statusline/renderer.go:591` shows that layout was removed and the counts now ride the repo segment. Documenting the new key on top of the old description would have left the page self-contradictory.

## `moai-goal` is not a duplicate

The two pages carry an explicit self-describing split — `utility-commands` is the verb/flag reference, `workflow-commands` is the concept and progression mode, each linking the other. Both kept; only the real defect was fixed (a `new_items` NEW badge rendering for a sidebar row that had no `title:` and no menu ref).

## Assets

Two infographic sets × 4 locales, generated and then **individually inspected**. The first pass produced invented API text inside the terminal illustrations (`send_msg 'hello'` and similar, none of which exist) and one garbled English variant; those five were regenerated with abstract bars instead of readable code. Palette reduction to 128 colors took them from 3-4 MB to 89-191 KB with no visible quality loss.

Three hand-authored SVGs mirror the README's mermaid blocks for docs-site use. The mermaid blocks stay, since GitHub renders them natively — this deviates from the `svg-infographic` skill's "never ship both" rule, so each SVG header names `README.ko.md` as the source of truth to bound the drift that creates.

## Verification

- README parity: 818 lines × 4, 12 H2 at identical line numbers (`40 153 244 308 414 512 574 629 686 753 777 806`), 65 H3 each
- All 37 image references resolve on disk; per-locale suffixes correct
- `hugo --printPathWarnings` warning-free; 161 files per locale, path diffs empty
- No mermaid `LR`/`RL`; no URL-blacklist hits
- No Go, template, or `.claude/` files touched — docs and assets only

## Known gaps, not in this PR

`cacheStrategy`, `moai-domain-design-dna`, and `MOAI_SESSION_PID` have zero docs-site coverage. Worth a follow-up card.

## Release state

`v3.1.1` is **not tagged yet** and its GitHub release is still a draft — that is `release-drafter` working as designed, not a fault. Tagging must go through `scripts/release.sh`, which stamps the provenance trailer `release.yml`'s gate requires; a hand-made tag is rejected and manual `gh release create` conflicts with GoReleaser. **This PR should merge before that tag is cut**, so the release ships with its documentation.

🗿 MoAI

## Summary by CodeRabbit

* **Documentation**
  * Updated README and changelog content across four locales for v3.1.1.
  * Added guidance for home-directory relocation, cleanup, disk-usage diagnostics, and retention settings.
  * Documented unattended Kanban `/loop` behavior, enhanced statusline options, and cross-session configuration.
  * Expanded CLI references for todo management, memory, tokens, cleanup, and ast-grep.
  * Updated skill inventories and added navigation for new workflow, CLI, and advanced pages.
  * Added planning reports and inventories to track documentation coverage and consistency.


GoosLab:
## Follow-up audit — four stale claims corrected, one of them mine

A full sweep of every v3.1.1 surface against `README × 4` and `docs-site × 4` turned up 9 gaps and **6 stale claims**. Stale claims outrank gaps — a gap leaves a reader with nothing, a stale claim sends them the wrong way — so those were fixed first, in `023562c3e` and `ea76a118b`.

### The one this PR introduced

The v3.1.1 highlights added earlier in this PR asserted that `cache.yaml` **governs** cache-breakpoint lifetimes. It does not. `internal/config/cache_config.go` records that the `cache_control` injector was never reachable from production and was removed, that `LoadCacheConfig` has zero callers (`grep -rn 'LoadCacheConfig' --include='*.go' . | grep -v _test` returns only the defining file), and that prompt caching is performed by Claude Code itself — with an `@MX:CEILING` warning that the file *"misleads as soon as a user expects the values to change behaviour"*. The paragraph did exactly that, in four locales.

Removed from the highlights. The config-table row stays — the file ships and shows up in the `moai web` settings editor — but now says plainly that nothing reads the values. Its old placement claim (`session_ttl` on session-start context, `spec_ttl` on SPEC bodies) came from the same removed injector and went with it.

Root cause worth recording: the change inventory found a new config file and I classified it as a new capability without checking whether anything consumed it. File existence read as feature existence.

### The other three

| Claim | Reality |
|---|---|
| `kanban-mode.md:237` — the factory lead "polls the backlog queue" | `internal/hook/session_start_factory.go:98` states in capitals that queue polling is **deliberately not taught** to the factory lead; doing so hands it a second, conflicting polling protocol. The sentence also contradicted itself — lead polls, *and* routes what the operator picked. `launchers.md:36` and `manager-lead.md:35` were already correct, so the docs disagreed with each other too. |
| `moai-todo.md:77` — `moai todo done` is the only way to remove an item, and nothing sets `dropped` | `drop`, `undrop`, `edit`, `move` all exist (`internal/cli/todo_drop.go:66,149`, `todo_edit_move.go:45,110`). The page was actively telling users a command they need does not exist. |
| `skill-guide.md` — 32 skills, 6 domain | 34 and 7. The READMEs were corrected earlier in this PR, so the two surfaces had started disagreeing. Derived token arithmetic updated with the totals; the core/optional split was also wrong (19 → 21). |

`config-sections.md` additionally documented `statusline.yaml` without its new `forge` key and had no `crosssession.yaml` section at all. Both added — with `isolate_machines: false` (no approval before a message leaves this machine) given its own row rather than buried in prose.

### Deliberately not done

- **No `cache.yaml` docs-site page.** Nothing reads those values; a page would multiply the misleading claim across four more locales. `cost-optimization/prompt-caching.md` correctly describes Claude-Code-side caching and was left alone.
- **`MOAI_SESSION_PID` is not a gap.** `internal/cli/launch_session_pid.go:12` says a hook must never set it — it is stamped by the launcher. No user configures it, so it does not belong in user docs.

### Still open

- **`moai migrate profiles` is undocumented** in all 8 files. `internal/cli/migrate_profiles.go:362` describes the failure it repairs: anyone who has run `moai cc -p <name>` has project memory split in two, and *"a session launched one way cannot recall what a session launched the other way learned"*. Users hitting this have no way to discover the fix. Highest-value remaining gap.
- **Cross-session messaging has no docs-site page** in any locale, only doctrine in `.claude/rules/`.
- **`internal/template/templates/.moai/config/sections/cache.yaml` contradicts itself** — line 12 comments "Ships disabled; flip to true to opt in", line 13 is `enabled: true`, and the web console string matches the comment. A code/template question rather than a docs one, so untouched here.
- Five lower-priority Factory Mode facts (`/clear` between cards, class collapse in-lane, release-branch integration discipline, the free-slot notice line, the lane join line) remain undocumented on both surfaces.

Verification unchanged and re-run: READMEs 816 lines × 4 with identical H2 offsets, `hugo --printPathWarnings` warning-free, 161 files per locale with empty path diffs.

🗿 MoAI

PR #1590, opened 2026-08-20
fix(release): stop publishing the raw commit list as the release body

GoosLab:
## Why

`.goreleaser.yml` carries `changelog.use: github`, so GoReleaser writes a commit-list body at publish time. Nothing in the pipeline overwrites it — that has to happen by hand afterwards, and the release is live with the auto body until it does.

v3.1.1 shows the gap concretely: published at `09:15:39`, body replaced at `09:31:03`. For fifteen minutes the public release page carried the generated commit list. The manual step ran, late; it is not wired into anything that guarantees it runs, or runs promptly.

The same phase also blocked on a per-version Korean notes file that had to exist first, which put the English body behind a translation step nobody was performing.

## What changes

**`.goreleaser.yml`** — `changelog.disable: true`. The release now arrives with an empty body instead of a competing one, so a missed follow-up leaves a blank page rather than a wall of commit subjects.

**Release harness Phase 7** — fills that body via `gh release edit --notes-file`, rewritten from *extraction* to *composition*, and moved to run immediately after the workflow completes rather than at leisure. Neither available source is safe to paste verbatim, for opposite reasons: the auto commit list buries what changed, and the `## [X.Y.Z]` CHANGELOG section is an internal development record (SPEC ids, AC numbers, commit SHAs, coverage figures). The notes are written from the CHANGELOG's `### Summary`, for someone deciding whether to upgrade. Phase 4 gains a matching obligation: `### Summary` is load-bearing, so a theme it omits is a Phase 4 defect rather than something Phase 7 papers over.

**Korean release notes retired** — `release.sh` no longer appends `.moai/release-notes/vX.Y.Z.ko.md`, and the four docs-site changelog pages no longer claim GitHub Releases carries a Korean body. The existing `v3.1.0.ko.md` stays; the retirement is forward-looking.

**docs-site changelog pages restated as pointers, not mirrors** — the per-release obligation there is a check that the pages still describe reality, not a copy into four locales that drifts silently.

## Verification

- `bash -n scripts/release.sh` — clean
- `grep -c 'release-notes\|ko.md' scripts/release.sh` → `0` (removal complete, no dangling reference)
- `.goreleaser.yml` line 47 → `disable: true`
- v3.1.1 publish/update timestamps via `gh api repos/modu-ai/moai-adk/releases/tags/v3.1.1` — the 15-minute exposure window above
- All four locales move in this commit (4-locale parity)

The size of v3.1.1's pre-replacement body is **not** verifiable — GitHub keeps no history of a replaced release body, so any figure for it would be an unobserved claim. What is observable is the timestamp gap.

`goreleaser check` reports a **pre-existing** deprecation unrelated to this change: `archives.format_overrides.format` should be `formats` in GoReleaser v2. It is a warning, not an error — v3.1.1 built successfully with it — and the action pins `version: "~> v2"` (floating), so a future v2 minor removing the key would break the build. Left out of scope deliberately; worth its own card.

No template mirror needed — `.claude/agents/harness/hns-release-specialist.md` is dev-only and has no counterpart under `internal/template/templates/`.

🗿 MoAI

## Summary by CodeRabbit

- **New Features**
  - GitHub Releases now provide consistently composed English release notes with clear summaries.
  - Published releases continue to include downloadable build assets.

- **Bug Fixes**
  - Release notes are now populated reliably instead of being obscured by low-level commit details.
  - Release links and descriptions across supported documentation languages now accurately identify English-only release content.

- **Documentation**
  - Updated release guidance and changelog references to reflect the streamlined English-only release format.


GoosLab:
Both Major findings addressed in `ff75553ad`.

**Release Drafter contract** — the template told maintainers to write `CHANGELOG.md` in English and Korean, which is the convention this path retires. It now states the English-only rule and names the `### Summary` section the harness composes the published notes from.

**Workflow wait** — the lookup resolves the tag to its commit (`git rev-parse "$VERSION^{commit}"`) and filters with `gh run list --commit`, so a re-run of an earlier tag or a concurrent release can no longer be watched in place of this one, nor its verdict reported as this release's. It polls for the run rather than reading once five seconds after the push (bounded at 180s, since the run does not exist the instant the tag lands). The wait itself is now a poll with a 45-minute ceiling instead of an unbounded `gh run watch`; past the deadline the script detaches and prints how to resume watching. The same instruction is written into harness Phase 7 so the script and the doc cannot drift.

Verified: `bash -n scripts/release.sh` clean, `.github/release-drafter.yml` parses as YAML.

@coderabbitai review

GoosLab:
The `Merge Risk` summary on `ff755` says the automation "still does not identify the workflow run for the release commit". That sentence quotes an inline finding written against `2773a3ec0` — the commit **before** the fix. The fix is `ff75553ad`.

Requirement against current `scripts/release.sh`:

| Asked for | In the file now |
|---|---|
| Resolve `vX.Y.Z` to its commit | `TAG_SHA="$(git rev-parse "${VERSION}^{commit}")"` |
| `gh run list --commit "$sha" --limit 1` | exactly that, with `--jq '.[0].databaseId // empty'` |
| Stop at a deadline no later than 45 minutes | `WATCH_DEADLINE=$(( $(date +%s) + 45 * 60 ))`, then detach with a resume hint |
| Apply the same commit filter in `scripts/release.sh` | this *is* that file |

One deviation, deliberate: the watch is a poll rather than `gh run watch --exit-status`. `gh run watch` blocks until the run ends and carries no deadline of its own, so it cannot satisfy the 45-minute bound in the same breath — wrapping it in `timeout(1)` would not be portable either, since coreutils is not present by default on macOS. The poll reads `gh run view --json status` every 30s and breaks on `completed`, which gives the same verdict with a bound that actually holds. The existing `conclusion` switch below it is unchanged, so a `failure` still exits non-zero.

Run-lookup is also bounded (180s) rather than a single read five seconds after the push — the run does not exist the instant the tag lands, and a run that never appears is now a reported failure rather than an empty `RUN_ID` silently skipping the wait.

The same instruction is written into harness Phase 7 (`hns-release-specialist.md`), so the script and the doc cannot drift apart.

The remaining Minor wording finding on `.github/release-drafter.yml:84` is worth taking; I will fold it in.

PR #1592, opened 2026-08-21
feat(gate): add a typecheck axis so a broken build cannot pass

GoosLab:
## Why

A consumer repository shipped a type error through a green `moai gate`.

The cause is a gap, not a bug: that repo uses **biome** rather than eslint, so the lint step skipped for want of an eslint config — and the Node toolchain entry has **no vet steps at all**, leaving only the test step, which never type-checks. Node was the one type-unsafe hole in the matrix:

| Language | Where type safety lives |
|---|---|
| Python | mypy, on the **lint** axis |
| Dart | analyze, on the **vet** axis |
| Go | `go vet` type-checks as it runs |
| Compiled languages | the test step compiles first |
| **Node** | **nowhere** |

A temporary workaround already landed on the consumer side. This is the root repair.

## What

A `typecheck` axis, running **between vet and lint** — a type error surfaces before the slower style pass, and a project whose linter config is absent still gets a correctness gate.

Node resolves in three tiers:

1. `gate.typecheck.command` — works for **any** language; how a Python or Go project opts in
2. `package.json` `scripts.typecheck` → `npm run typecheck`
3. `tsconfig.json` present → `npx --no-install tsc --noEmit`

**A solution-style tsconfig is refused rather than run.** `files: []` plus `references` type-checks nothing and exits 0, so accepting it would rebuild the very blind spot being closed. Tier 2 deliberately outranks that check, so a monorepo root delegating to turbo is not penalised for having one.

## Skipping is not failing, but it is never silent

The silence is what let the broken build through. A language with no default now says so, and says how to supply one:

```
typecheck: skipped (no scripts.typecheck and no tsconfig.json; set gate.typecheck.command to enable one)
```

Tool availability stays fail-open. Only the visibility changes.

### A second silence, found on the way

`passReason` **assigned** rather than accumulated, so the last notice erased the earlier ones — this axis's own notice would have vanished whenever ast-grep also had something to report. Vet and lint notices were dropped outright. They accumulate now.

## Config

```yaml
gate:
  timeouts:
    typecheck: 300   # a monorepo type-check routinely outlasts the test budget
  typecheck:
    enabled: true
    command: ""      # any language; split on whitespace, no shell metacharacters
```

Both `gate.yaml` files updated; the three keys are registered in the shipped-key triage inventory as wired, with the gate mapping as their reader. `CONFIG_STRUCT_YAML_MISMATCH` passes.

## Verification

TDD, RED first. 14 new tests, including a **pass counterpart** for the blocking test — a block-only assertion is satisfied by a gate that always fails.

Real-binary demo against three fixtures:

| Fixture | Result |
|---|---|
| failing typecheck | gate **blocks**, prints the type error verbatim |
| passing typecheck | gate proceeds past the axis |
| no script, no tsconfig | gate **passes**, reports the skip alongside ast-grep's |

`golangci-lint` 0 issues · `go build ./...` clean · `make build` clean, catalog unchanged.

## Known limitation (pre-existing, out of scope)

`runStep` does not set `cmd.Dir`, so a step **resolves** against the project dir but **executes** in the process cwd. Real use calls `moai gate` at the repo root, where the two match; this axis inherits the looseness rather than introducing it.

## Not verified

- Full suite is CI's call, not run locally.
- `tsc` itself was never executed — tier 3 is tested as far as constructing the command; the demos all ran tier 2.
- Consumer auto-firing is inferred from its `scripts.typecheck`, and needs measuring there after release.

🗿 MoAI

## Summary by CodeRabbit

* **New Features**
  * Added a configurable typecheck step to the quality gate, enabled by default.
  * Supports custom commands, package scripts, and TypeScript fallback detection.
  * Typechecking runs between vet and lint with a 300-second default timeout.
  * Reports successful, skipped, and unavailable checks without silently ignoring them.
  * Skipped checks remain non-blocking, while failed typechecks block the gate.
* **Documentation**
  * Added configuration guidance for typecheck behavior, precedence, overrides, and TypeScript project handling.

### DIFF
--- a/.moai/config/sections/gate.yaml
+++ b/.moai/config/sections/gate.yaml
@@ -17,6 +17,30 @@ gate:
     vet: 30
     lint: 60
     test: 120
+    # A type-check over a large monorepo routinely outlasts the test budget,
+    # so it gets its own.
+    typecheck: 300
+  # The typecheck axis runs between vet and lint, so a type error surfaces
+  # before the slower style pass and a project whose linter config is absent
+  # still gets a correctness gate.
+  #
+  # Node resolves in three tiers: `command` below (any language), then
+  # package.json scripts.typecheck, then `npx --no-install tsc --noEmit` when a
+  # tsconfig.json is present. A solution-style tsconfig (files: [] plus
+  # references) is refused rather than run — it type-checks nothing and would
+  # pass vacuously.
+  #
+  # Every outcome is reported. A skip is not a failure, but it is never
+  # silent: a language with no default prints how to supply one.
+  typecheck:
+    enabled: true
+    # Overrides the language default and enables the axis for ANY language,
+    # e.g. "mypy --strict ." or "npm run typecheck".
+    #
+    # Split on whitespace and executed directly, so shell metacharacters are
+    # NOT interpreted — put a pipeline or a redirect in a script and name the
+    # script here.
+    command: ""
   ast_grep_gate:
     enabled: true
     # rules_dir is the single source of truth for where this checkout's
--- a/.moai/reports/t172/verdict.md
+++ b/.moai/reports/t172/verdict.md
@@ -93,8 +93,8 @@ FAIL	github.com/modu-ai/moai-adk/internal/hook/quality [build failed]
 ### 5.2 대조절 포함 테스트 14건
 
 차단(RED)만 있으면 "항상 실패하는 게이트"로도 만족된다. 그래서 **통과 대조절**을 같이 뒀다:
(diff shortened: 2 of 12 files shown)


GoosLab:
Manual merge override (operator-approved 2026-08-22, factory tk4ntu lead). CodeRabbit Merge Risk 🟡 Moderate — 3 findings accepted as documented bounded limitations: (1) BOM/solution-config variant TS bypass: solution-style already blocked by design; BOM variants remain a known edge, users can force tier (a) via gate.typecheck.command. (2) runStep cmd.Dir unspecified: explicitly out of card scope, consistent with existing gate steps. (3) 300s typecheck timeout default: config value, adjustable via gate.timeouts.typecheck. All CI checks green incl. Race Test (9m33s) and Template Neutrality Audit.

Issue #1265, opened 2026-08-01
[Bug] gate.yaml is never loaded in v3 — quality gate cannot be disabled (Loader.Load lacks loadGateSection; #786 recurrence)


wert-ysjang:
## Describe the Bug

`.moai/config/sections/gate.yaml` is **never read** by the v3 binary, so the pre-commit quality gate cannot be disabled or tuned by any supported means.

`Loader.Load()` (`internal/config/loader.go:31-101` at tag `v3.0.1`) wires loaders for user/language/quality/git-convention/git-strategy/llm/ralph/state/workflow/statusline/research/feedback/handoff/archive/constitution/context/interview/design — but there is **no `loadGateSection`**. As a result `cfg.Gate` is always `NewDefaultGateConfig()` (`enabled: true`, `defaults.go:277-288`), regardless of what gate.yaml says. There is also:

- no env override for the gate (`applyEnvOverrides` in `manager.go:392-405` only covers dev-mode/log keys), and
- no config path that populates `quality.GateConfig.DisabledSteps` (the #667 Fix 3 knob is defined but never set from any source).

Meanwhile the scaffolded `gate.yaml` explicitly documents `enabled: "Set to false to skip all checks"` and `skip_tests`, and `internal/hook/pre_tool.go:545-587` (`loadGateConfig`) faithfully maps `cfg.Gate` — the struct, defaults, and mapping all exist; only the file→struct loader is missing.

Two contributing gaps that let this ship:

1. `internal/config/audit_registry.go:37` registers `"gate": "GateConfig"` in `yamlToStructRegistry`, whose contract says every entry "has a loader" — but the parity test (`audit_test.go`) only checks yaml-file ↔ struct existence, not loader wiring.
2. `internal/defs/dirs.go:242` lists `.moai/config/sections/gate.yaml` as deprecated (removal v3.0.0) with no documented replacement, yet the template still ships the file.

This is a recurrence of the #786 class ("ignoring `gate.enabled: false`", v2.14.0, closed COMPLETED) — in v3 the ignore is structural: no code path reads the file.

## To Reproduce

```bash
mkdir lab && cd lab && git init -q
echo pytest > requirements.txt              # any Python marker
mkdir -p .moai/config/sections
printf 'gate:\n  enabled: false\n  skip_tests: true\n' > .moai/config/sections/gate.yaml
echo '{"toolName":"Bash","toolInput":{"command":"git commit -m test"}}' | moai hook pre-tool
```

[...]

PR #1625, opened 2026-08-23
fix(hook): take the quality gate out of PreToolUse's 10s budget (t214)

GoosLab:
> **Stacked on #1624.** That PR's `cmd.Dir` fix is a prerequisite, not a convenience: before it, `runStep` ran in the caller's cwd, so a test that exercises this gate re-executes the suite invoking it. Merge #1624 first; this branch already carries it as a merge commit.

## The defect

A `git commit` command ran the full vet+lint+test quality gate **inline inside the PreToolUse hook** and denied the tool call on failure.

```
gate step ceilings   vet 30s · lint 60s · test 120s
hook entry timeout   10s
measured             30,033 / 30,016 / 30,020 ms  →  permissionDecision: deny
deny reason          "quality gate timed out: go test exceeded 2m0s"   ← the 30s vet ceiling tripped, not this
```

On any repository large enough to need the gate, Claude Code kills the hook long before the gate's own ceiling. The user waits 10s and the commit path is unreliable. **A gate that cannot finish inside its budget has produced no evidence, so it must not deny on that basis** — and the reason it denied with was wrong about which budget tripped.

This session was masked by `MOAI_AUTONOMY_TIER=fully-autonomous`, which appears in neither settings file. A default session paid it in full.

## The gate moves to where the code already said it belonged

`moai init` / `moai update` install a git `pre-commit` hook that shells out to `moai gate` (`internal/cli/hook_install_precommit.go`): the same 16-language gate, no 10s ceiling, a clean commit abort instead of a tool-permission deny, and a documented `SKIP_MOAI_PRECOMMIT=1` bypass. The comments there already call it the *"relocated heavy quality gate"* — the PreToolUse copy was the half of the relocation that never left.

**Why not cap it at 8s and downgrade the timeout to allow-with-warning.** The measurements rule that out: a trivial module completes in 1.1s, this repository takes 30s. An 8s budget therefore times out *every time* on a real repository — that does not shorten the block from 30s to 8s, it makes 8s of useless blocking permanent.

## Fallout, handled rather than deleted

**Five typecheck-adjacent frames.** The ast-grep skip-reason test drove its assertion through three frames, the last being the handler running the gate inline. It now stops at `QualityGate.Run`, whose surviving consumer is `moai gate` (`runGate` prints a non-empty pass-path output to stderr). The propagation under test is unchanged; only the frame that consumed it is gone.

**`loadGateConfig` / `firstLine` are left in place** despite having no production caller. Their tests are the only regression guard for issue #1265's `disabled_steps` mapping (`grep DisabledSteps internal/cli/*_test.go` → 0 hits, so the CLI-side mirror `mapConfigGateToQuality` has none). Deleting them would remove the guard for a real past defect; moving the guard to the mirror and then deleting is follow-up work, not this PR.

## `status-transition-ownership.sh` → async

The one PostToolUse entry still on the blocking path, for a pure advisory audit-log append. Its own comments state it is advisory, that `exit 2` blocking is *"reserved for future enforcement"*, and that nothing reads its stdout. All three template entries gain `"async": true`.

**Premise correction on the audit's 17s figure.** That number is real but measures the local rendered `.claude/settings.json`, which predates the path scoping the template already carries:

| | local rendered `settings.json` | template `settings.json.tmpl` |
|---|---|---|
| entries | 1 | 3 (Write / Edit / MultiEdit) |
| path scope | **none** — fires on every Write/Edit | `"if": "<Tool>(**/.moai/specs/**)"` |
| `async` | absent | absent → **added here** |

On an updated project it fires per SPEC-artifact write, not per Write/Edit. The remaining real defect was the missing `async`.

## Verification

The reproduction test uses a tiny fixture, not a 30s one: whether the gate runs is the question, not how long it takes. The fixture carries a printf mismatch `go vet` reliably flags, so a gate still running inline denies on it.

```
RED    --- FAIL  PreToolUse denied a plain `git commit` against a vet-failing fixture (0.15s)
GREEN  --- PASS  TestPreTool_GitCommitDoesNotRunTheQualityGate (0.00s)

go test ./internal/hook/ -count=1          ok  57.168s
go test ./internal/template/... -count=1   ok  33.430s · ok 1.161s
make build                                 templates re-embedded, build OK
go build ./internal/...                    clean
```

## Not in this PR — and why

The card also carried F-2b (merge `handle-security-scan.sh` into `handle-post-tool.sh`) and F-4 (drop the PreToolUse ast-grep scan, keeping the async PostToolUse one). Both change **which security scan runs and when** — F-4 removes the ability to deny before a write lands. Bundling those into a latency PR invites a reviewer to read a security-surface change as a performance tweak. They are carded separately, stacked behind this one, and that card declares itself a security-surface change in its own title.

## Requirement retirement record — SPEC-STOPCHAIN-TRIM-001 REQ-005

Recorded here rather than by editing the SPEC. A SPEC is a record of what landed at the time it landed; rewriting it to match current behaviour would make that record false. REQ-005 really did build a tier-aware commit gate, and it really did work that way. This section is where the question "what happened to REQ-005?" gets its answer, in the commit history.

| | |
|---|---|
| **What is retired** | The tier-aware synchronous commit gate in `internal/hook/pre_tool.go`, and with it the `MOAI_AUTONOMY_TIER` switch that turned it off at `automatic` / `fully-autonomous`. |
| **Why** | Its budget exceeded the hook's by 21×, so on any real repository it was killed before finishing and denied on evidence it never produced. It was not a gate that could be tuned; it was a gate that could not run. |
| **What survives** | REQ-005's intent — a commit at semi-auto is verified before it happens. The git `pre-commit` hook enforces that unconditionally, with no ceiling and a clean abort. |
| **What is genuinely lost** | The ability to disable commit verification *by autonomy tier*. Nothing replaces that axis. |
| **The remaining bypass** | `SKIP_MOAI_PRECOMMIT=1`, which is per-invocation and tier-blind — a different axis, not a rename of the old one. |

An unattended run that previously relied on `fully-autonomous` to skip verification now pays the pre-commit gate on every commit unless it sets that variable. That is a behaviour change for unattended sessions and is stated here so it is not discovered as a surprise.

## Residual risk

**A user whose pre-commit hook was never installed** (the installer refuses to overwrite an unmarked existing hook) now has no commit gate. They had no effective one before either — the 30s gate died at 10s and its deny carried no evidence — so this removes an absent protection, not a working one.

**The `async: true` runtime effect is not observed here.** Whether Claude Code honours the key is runtime behaviour; the evidence is that sibling entries in the same file already carry it, not a fired hook.

Evidence: `.moai/reports/t214/verdict.md`

🗿 MoAI

## Summary by CodeRabbit

- **New Features**
  - Added automatic status-transition ownership checks after editing specification files through Edit and MultiEdit operations.
  - Existing checks continue to run after Write operations.

- **Bug Fixes**
  - Git commits are no longer blocked by an in-tool synchronous quality gate when the pre-commit hook is responsible for verification.
  - Destructive command safeguards remain active.

- **Tests**
  - Added coverage confirming commits proceed without the removed synchronous gate.
  - Improved validation of quality-check skip reasons.


GoosLab:
@coderabbitai review — retriggering: the previous head was review-rate-limited and an empty commit did not produce a review invocation.

PR #1627, opened 2026-08-24
docs(kanban): pre-dispatch PR cross-check + card id in PR titles (t210)

GoosLab:
SPEC-KANBAN-PR-CARD-TRACEABILITY-001 (Tier S, doctrine only — no Go code).

Card: t210. Sibling: SPEC-KANBAN-QUEUE-PR-SYNC-001 (the tooling), which lands second.

## What this changes

Two [HARD] clauses in `.claude/rules/moai/workflow/kanban-dispatch.md`, mirrored into the distributed template.

**1. The pre-dispatch PR cross-check** (§ Entry into the board is an operator act). Before dispatching a card out of `backlog`, the lead reads that card's pull-request and landed state and reports what it read in the same turn.

The measured failure was not a lead that looked and misread — it was that nothing required looking. Five cards sat `queued` while each carried an open PR; one sat `queued` while its fix was already an ancestor of `origin/main`, discovered only after a lane had started. That cost a full lane, and it is the only sub-case in the record with a quantified cost.

**2. The clause reports; it never vetoes.** Where a card turns out to carry an open PR or to be already landed, the lead surfaces that and the operator **confirms or withdraws** it. This wording is the whole control: the operator has already picked the card, and "promotion is the operator's act, always" — a lead that refuses a picked card on its own authority has overridden an operator act rather than informing one. No mechanical check can tell the two readings apart after the fact, which is why it is a separate clause with its own acceptance criterion.

**3. A card-delivering PR title MUST carry the card id** (§ Isolation), stated next to the branch-naming rule it reconciles with. This does not contradict the [HARD] rule excluding the card id from branch names: the branch name is read by a human scanning `git branch` and wants a descriptive slug; the PR title is read by a machine and wants the id. Traceability now rests on four carriers rather than three — dispatch `card:` field, commit message, evidence path, and the PR title, the only machine-readable one. A reader meeting both [HARD] clauses cold will suspect a conflict, so the doctrine says outright there is none.

The obligation binds card-delivering PRs only; release, batch, and maintenance PRs carry no card id and no obligation. It binds PRs opened after it lands — nothing is retitled.

## Why doctrine lands before the tooling

The title convention is what raises the title carrier from 64% recall toward full coverage, which turns the sibling SPEC's resolver from a heuristic into a lookup. **The fix for ambiguous parsing is a naming convention, not a smarter parser.**

[...]

PR #1614, opened 2026-08-23
fix(todo): refuse a mistyped verb that addresses a card id (t203)

GoosLab:
Card t203 (Class C — discriminator direction confirmed by the lead). Closes #1597.

## What was wrong

`moai todo pick t151` created a card named `pick t151` instead of erroring. The t69 fallthrough reads any two-or-more-word invocation as natural language, and a mistyped dispatch verb is exactly that shape — so a typo became a silent queue mutation. The lead hit it in operation.

## The discriminator

The fallthrough stays: it is what makes `moai todo fix the flaky gate` work, and t69 accepted mistyped-verb-plus-words as its price. This narrows the one shape where that price is highest.

**Refused**: a verb-shaped first token (`^[a-z][a-z-]{1,15}$`) followed by a card id in second position (`^t\d+$`).

Three deliberate narrowings, each protecting a real card:

| Choice | Why |
|---|---|
| Registered verbs never reach the check | Cobra routes them to their subcommands, so anything at the parent is already an unregistered word |
| Id must be the explicit `t151` form, not the looser `151` the verbs accept | `moai todo fix 3 flaky tests` must still add |
| Only the **second** token is checked | `moai todo fix the drift found in t151` must still add |

The error names the registered verbs and the escape hatch, and the verb list is derived from `cmd.Commands()` rather than copied — so it cannot drift as verbs are added. (13 are registered today; the issue lists 6 and the card 9.)

## Evidence

Built binary, isolated queue:

```
$ moai todo pick t151
ERROR  Todo: "pick" is not a todo verb and "t151" is a card id — refusing to create a card named "pick t151".
       Known verbs: add, analyze, done, drop, edit, list, move, next, relate, undrop, unpick, unrelate, why
       To add this text as a card anyway: moai todo add "pick t151".
EXIT=1

$ moai todo fix the drift found in t151   → t1 1        (exit 0)
$ moai todo next t1                        → picked t1  (exit 0)
```

**Falsification probe**: neutering the guard turns `TestTodoMistypedVerbWithCardIDErrors` red; reverting it turns it green. The test observes the guard rather than a string.

| Command | Result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./internal/cli/` | exit 0 |
| `GOOS=windows go vet ./internal/cli/` | exit 0 |
| `go test ./internal/cli/... -count=1` | all packages ok (`internal/cli` 344.7s) |
| `golangci-lint run ./internal/cli/...` | `0 issues.` |

## Residual risk

[...]

PR #1628, opened 2026-08-24
feat(kanban): moai todo pr — read-only card-to-PR and landed link (t210)

GoosLab:
SPEC-KANBAN-QUEUE-PR-SYNC-001 (Tier M). Card: t210.

Sibling: **#1627** (`SPEC-KANBAN-PR-CARD-TRACEABILITY-001`, the doctrine), which lands first — its PR-title convention is what makes this resolver a lookup rather than a heuristic. The two PRs share the four plan-phase commits by SHA, so once #1627 merges they drop out of this diff.

## What this adds

`moai todo pr [<id>]` — answers, for every card in the queue, whether an open pull request already delivers it or whether its work has already landed on `origin/main`. Those are the two divergences that put five cards in the queue with open PRs and cost a full lane on a card that was already merged.

## Two questions, two carriers — deliberately not conflated

| | Question | Carrier | Cost |
|---|---|---|---|
| **Q1 — attribution** | which card does this OPEN pull request deliver? | PR title, then PR body | one `gh pr list` |
| **Q2 — landed** | is this card's work already in `origin/main`? | landed commit messages | local `git log` |

Measured across 11 open PRs: the **title** carrier is precise (7/7 — every token present is the delivering card) but incomplete (64% recall); the **body** carrier has full recall and poor precision (#1614 carries 5 tokens for 1 card); **commit messages** are unusable for Q1 — #1600 carries a dozen-plus tokens and its own delivering card is not among them, because a branch that merges the release branch inherits every other card's commits.

That noise is **harmless for Q2**, and the asymmetry is the design. Q1 asks the commit set to name a card exclusively; an inherited commit breaks that. Q2 asks only whether `origin/main` names the card at all — and a commit that rode in on another branch is still genuinely landed. Q2 is the only carrier that reaches the card whose fix merged inside a batch PR carrying no token for it.

Four outcomes, distinguishable by kind alone: `linked` (confidence `exact` | `inferred`), `ambiguous` (every candidate listed, none chosen), `landed`, `no-link`.

## It observes; it never writes

No card field, no `findings[]` entry, no cache, no lock, no timestamp. The test digests **the whole queue directory** (every path + SHA-256) rather than `backlog.json` alone — a sidecar, a lock taken and released, or an mtime touch on a neighbour all pass a single-file check. Asserted on the linked, ambiguous, landed, and fail-open paths.

Auto-updating card state was considered and rejected: it is the queue mutation the dispatch doctrine forbids by name, it would make the queue a derived artifact with no operator provenance, and it fails asymmetrically — a wrong "this card is done" is contradicted by no later signal.

## Two guards worth reading

**The `-E` trap.** `\b` is not POSIX ERE and git fails **silently** under `-E`, returning empty — indistinguishable from "not landed". An `-E` regression would report every card as not-landed and pass any check that only looks for a clean result. The landed query mandates `--perl-regexp` via a named constant, and the suite carries a positive control (fails if the query returns nothing for a known-landed card), a negative control, and an `-E` tripwire built from the implementation's own argv.

**One `gh` query, not one per card.** `gh pr list` costs 0.878s, essentially all round-trip. `moai todo list` is the queue's cheapest read — every operator glance and the foreman loop — so the link is a **dedicated verb**, not a column, and no `--pr` flag is in scope. The census asserts exactly one `gh` process at queue lengths **3 and 10** (invariance, not coincidence — a one-card fixture cannot distinguish a batched query from a per-card loop) and that its argv is a single `pr list --state open`, not a per-card `pr view`. Without that criterion, a per-card implementation passes every other one while costing 0.878s × queue length.

## Fail-open

`gh` absent, unauthenticated, or offline: the link column renders blank (present, not omitted), the degradation is a `note:` on stderr rather than an error, the exit code is 0, and the landed check keeps running — it is local git.

## Verification

All 14 acceptance criteria pass. `go vet` clean, `golangci-lint run` 0 issues, `internal/kanban` 85.9% coverage (new resolver functions 100% except one unreachable defensive branch), `internal/cli` green (436s), template neutrality suite green, `make build` run. Full-suite verdict is CI's on this head.

Fixtures are **transcribed from the pinned measurement block**, never re-fetched — the live open set has already grown since it was taken, and a test that queries GitHub is a non-deterministic assertion about somebody else's repository state.

## Gaps — not claimed

- **AC-013's reviewer-judgement half** (does the mirrored doc row describe the verb accurately?) has no command and is recorded as an outstanding judgement.
- **The verb has not run against live GitHub.** Every `gh`/`git` call goes through an injected seam, so the real payload shape is verified against a transcription of a measured payload, not a fresh response.
- **Merged-PR attribution stays unscored** — Q1 queries `--state open` only. The t199 case is covered by Q2, which reaches it where a merged-PR extension measurably does not.
- **A card with both an open PR and a landed commit reports `linked` only**, masking the landed fact. That is REQ-1.9's specified behaviour and a known open finding; resolving it needs a seventeenth requirement, and this SPEC sits at exactly 16/16 where the answer is a tier-up or a split, never a quiet addition. No blocker was needed — none was written.

Evidence: `.moai/specs/SPEC-KANBAN-QUEUE-PR-SYNC-001/progress.md` §E.2 / §E.3.

🗿 MoAI

## Summary by CodeRabbit

- **New Features**
  - Added `moai todo pr [<id>]` to report pull-request and landed-work status for queued cards.
  - Displays linked, inferred, ambiguous, landed, and unlinked outcomes with confidence details.
  - Supports single-card filtering and JSON output.
  - Uses one batched lookup per invocation and keeps queue data unchanged.
  - Checks local history for landed work.
  - Continues gracefully when GitHub information is unavailable.

- **Documentation**
  - Added command usage guidance and behavior details for unavailable pull-request data.

## Added after the first review

- **Every subprocess is now bounded** (30s context). `gh` talks to the network and `git log` walks history; a read-only status verb that hangs is worse than one that degrades, and the fail-open path already renders a useful queue without either answer.
- **A saturated `gh pr list` page is reported, not hidden.** At the 100-record ceiling a card whose PR sits beyond it would resolve `no-link` or `landed` — wrong, and silent. Paging is not the fix (a second page is a second `gh` process, which the census forbids), so the ceiling is surfaced on stderr. Two tests pin it.
- **The git fixture is portable.** `GIT_CONFIG_GLOBAL=/dev/null` reads as empty config on Unix and names a nonexistent path on native Windows, where it would leave the real config in play.
- **The `§E.3` claim no longer opens with "all 14 pass"** while recording AC-013's judgement as outstanding — the criterion is partially discharged and the report says so first.

Review findings deliberately NOT acted on here: the `plan.md` M3 exit list, the `acceptance.md` fixture-excerpt label, and the REQ-1.1 confidence-scoping wording are SPEC **body** content, which run-phase may not edit under the ownership matrix — they are returned to the lead as a blocker list. The landed-fact-masked-by-`linked` finding is the known open item that needs a seventeenth requirement against a 16/16 budget.


GoosLab:
@coderabbitai review — retriggering: this head has been review-rate-limited since the batch quota freeze.

GoosLab:
@coderabbitai review — head moved with an empty commit; invoking fresh.

PR #1638, opened 2026-08-24
fix(worktree): repair the worktree reaper — three-valued merge detection, lock-aware anchor guard, clean --stale inventory (t209)

GoosLab:
## What this is

SPEC-WORKTREE-REAPER-001, card t209. The dispatching card asked for a worktree reaper. The plan-phase investigation found one already shipped, already enabled in this repository, and running on every `moai session register` / `moai session list` — preserving almost everything it looked at, blind to most of what it should guard, and unable to see a third of the trees at all. The SPEC was reframed as a repair, and the reframing is recorded in `spec.md` §B.1 because the card's premise differs from what was measured.

## The three milestones

**M1 — merge detection has three outcomes, not two.** `gh pr view` errors on the ordinary case of a merged PR whose head branch was deleted on the remote, and the sweep collapsed that error, an absent PR, and malformed JSON into one empty string it read as "not merged" — so the worktree was preserved, permanently, on every future sweep. Only the no-answer case now consults `git branch --merged origin/main`; a determinate `OPEN`/`CLOSED`/`DRAFT` ends the decision, so neither source overrides the other.

**M2 — the anchor guard reads the lock, not just the registry.** The registry's `cwd` is corrected only when the relocation hook fires: measured naming **1 of 5** live anchors, against **5 of 5** named by git's worktree lock reason. The decision moved to `internal/session`, reads the lock reason as its primary source, probes the pid it names for liveness, and unions the registry in as a supplementary input. It fails closed on every parse or probe failure. All three previously-blind sweeps now share it.

**M3 — `moai worktree clean --stale --json`** reports every non-protected registered worktree with the dirty, merge, anchor, and ignored predicates behind its keep reason, and removes nothing — it overrides `--yes` rather than combining with it. The base branch for `--merged-only` and `--stale` now defaults to `origin/main`; a local `main` behind the remote made the two sweeps disagree about the same tree.

## Audit trail

Plan-audit ran three iterations (FAIL 0.55 → FAIL 0.84 → **PASS 0.875**). Sync-audit ran three (FAIL 0.778 → FAIL 0.844 → **PASS 0.887**, threshold 0.85), and each failure was a real defect rather than a rubric artifact:

- **F1** the git fallback stripped `*` but not `+`, and git prefixes `+` on every branch checked out in a linked worktree — which is every candidate this sweep evaluates. Measured on this repository: **119 of 149** merged branches read as not merged, so M1's whole fallback path was inert. The parser is extracted and bound by a test that drives the seam *through* it, closing the blind spot that let the defect ship green.
- **F2** an unreadable `git worktree list --porcelain` silently dropped the authoritative lock source and degraded both `clean` sweeps to registry-only, with nothing in the output to say so. Both now fail closed and name the cause.
- **F3 / N1** `git status --porcelain` and non-forced `git worktree remove` **agree** in disregarding ignored files, so nothing stood between the sweep and a tree whose only content was `.claude/agent-memory/`. The guard now covers all three sweeps from one decision — allowlist included — rather than a second copy free to drift.
- **F4** the trailing-segment base guard the `origin/main` default made load-bearing had zero tests; `feature/main` not matching `origin/main` is now locked in.

Reports: `.moai/reports/t209/sync-audit{,-iter2,-iter3}.md`. Evidence: `progress.md` §E.1–§E.6.

## Verification

`go build ./...`, `go vet ./...`, `GOOS=windows go vet ./...` — exit 0. `golangci-lint` — 0 issues. `internal/session` and `internal/cli/worktree` green. `internal/cli` was measured green at `aa14918d7` (`ok … 768.952s`, exit 0, uncontended) and is cited as prior attributed evidence rather than re-run; **CI owns the full-suite verdict against this head.**

## What was NOT observed

- Windows runtime — compilation only.
- No sweep was run against a real repository: `clean` prunes shared worktree administrative state, and this repo holds 150+ registered trees.
- The `.claude/agent-memory/` preservation is a **stopgap**, recorded as such (REQ-WR-025). Draining those trees is the correct fix and is not in this change.
- `worktree done --auto` runs unattended after a PR merge without the ignored-content guard — deliberately outside it (single named target, not a sweep), and named for a follow-up card.

## Docs

CHANGELOG `[Unreleased]`. 8 docs-site files across 4 locales — two pages documented the old `--base` default, and `--json` was undocumented.

🗿 MoAI

## Summary by CodeRabbit

* **New Features**
  * Added non-destructive JSON inventory for stale worktrees via `moai worktree clean --stale --json`.
  * Worktree cleanup now defaults to `origin/main` and reports preservation reasons.
* **Bug Fixes**
  * Improved protection for locked, active, dirty, and ignored-content worktrees.
  * Added safer merge detection fallback and linked-worktree handling.
* **Documentation**
  * Updated worktree cleanup guidance and localized CLI references.
* **Chores**
  * Disabled automatic merged-worktree cleanup pending intentional review.


GoosLab:
Manual merge (operator override): CR Merge Risk is High (above the auto-merge threshold) — resolved as follows. (1) OD-1 (Major, classification-to-removal TOCTOU on the --stale path): fixed with a re-check immediately before removal, mutation-verified (TestCleanStale_RechecksIgnoredContentBeforeRemoval); the residual window between the second read and the syscall is documented in cr-response.md as narrowed-not-closed — no lower layer refuses removal for ignored files in this design. (2) OD-2 (Major, exit 2 on lock-source unreadable): rejected on verification — cmd/moai/main.go:13-22 exits 1 for every non-ExitCoder error, so the requested exit code is unreachable under current wiring; --json already reports anchored: undetermined + keep_reason: cause=lock-source-unreadable (REQ-WR-012); conflicts with REQ-WR-016. Follow-up card t231 filed (exit 2 + REQ-WR-016 revision + ExitCoder wiring). (3) Remaining actionable: 1 Minor — markdown fence language tags in a local report file, no behavior impact. CI 15/15 green on the merged head ca097280e.

Issue #1679, opened 2026-08-27
pre-commit fast-subset go vet fails on every Go-touching commit when the Go module lives below the repo top level (multi-root monorepo)


GoosLab:
# pre-commit fast-subset `go vet` fails on every Go-touching commit when the project's Go module lives below the repository top level

## Summary

The embedded pre-commit hook (observed on v3.1.3-rc.5) builds staged-package paths relative to the **repository top level** and runs a single batched `go vet` from CWD (= repo/worktree top):

```sh
PKGS="$(
    printf '%s\n' "$STAGED_GO" | while IFS= read -r f; do
        [ -n "$f" ] || continue
        printf './%s\n' "$(dirname "$f")"
    done | sort -u
)"
...
if ! go vet $BT_TAGS $PKGS >/dev/null 2>&1; then
```

When the project carries its Go module deep below the top — e.g. a bun/npm monorepo whose sole Go service is `apps/id/` — **every** commit that touches any `.go` file dies with

```
go: cannot find main module, but found .git/config in <repo-root>
	to create a module there, run:
	cd ../../.. && go mod init
```

even when the change is a comment typo fix and the package vets clean. Contributors are forced into permanent `SKIP_MOAI_PRECOMMIT=1`, which erodes the gate's meaning for commits where the check *would* carry information.

Single-module projects are unaffected (paths happen to resolve inside the one module), which is why this stayed latent.

## Reproduction

Layout: no `go.mod` at root; module present at `apps/id/go.mod`.

```console
$ printf '// probe\n' >> apps/id/config/loader.go && git add apps/id/config/loader.go
$ sh .git/hooks/pre-commit     # exit 1
[pre-commit] FAILED: go vet reported issues in the staged packages.
```

with the underlying cause (swallowed by the hook) being exactly the `cannot find main module` above. Also reproduced in an isolated three-repo sandbox covering multi-module, single-module, and outside-any-module layouts.

## Root cause

`go` resolves package arguments against CWD-upward module discovery. Path args rooted *above* their module boundary have no meaning to the toolchain; the old invocation only worked where the sole module happened to sit at the top level.

## Proposed repair

[...]

PR #1681, opened 2026-08-27
fix(quality): detect toolchain markers below project top and bind module roots

### LINKED ISSUE #1680: Heavy gate (moai gate) detects language markers only at the project top — nested modules get silently zero toolchain coverage (all 16 languages)
# Heavy gate (`moai gate`) detects language markers only at the project top — nested modules get silently zero toolchain coverage (all 16 languages)

> Companion issue to #1679 (which names this defect as "Adjacent observation 1" and owns the pre-commit fast-subset half). Filed separately because the component, the repair surface, and the verification matrix are different: this one is the **heavy gate** (`internal/hook/quality`), a Go-code change, not a shell template.

## Summary

`QualityGate.detectToolchain()` (`internal/hook/quality/gate.go`) tests each language's marker files at exactly one path — `filepath.Join(projectDir, marker)`. In a monorepo whose Go module (or Python project, Rust crate, …) lives **below** the repository top, the marker is never seen. The gate then takes the fail-open path:

```go
tc := g.detectToolchain()
if tc == nil {
    return true, gfNotice   // gate PASSES — zero checks ran
}
```

so every commit in such a repository passes the heavy gate with **zero** Go coverage, silently. This is the opposite failure character of #1679: that defect failed loudly and forced workarounds; this one passes quietly and skips the inspection it exists to perform.

The defect is **three-layered** — fixing only detection is not enough:

1. **Single-level marker scan** — `detectToolchain()` never descends below the project dir (`gate.go`, the `for i := range toolchains` / `filepath.Join(dir, marker)` loop).
2. **Single-toolchain assumption** — the loop returns on the *first* marker hit, so even with detection fixed, a repo with `package.json` at the root AND `go.mod` nested would still run only the Node toolchain.
3. **Project-top step execution** — `runStep()` sets `cmd.Dir = resolveQualityProjectDir(...)`. A naively-detected nested Go module would then run `go vet ./...` at the repo top, outside its module — flipping the failure from silent-pass to loud-fail (`go: cannot find main module`, the very failure #1679 fixed for the hook). The module root must be bound into step execution.

## Reproduction (measured, v3.1.3-rc.5 installed binary)

Sandbox mirroring the shape that triggers it in practice (our `mo.ai.kr` monorepo carries its sole Go service at `apps/id/`, no `go.mod` at top):

```
sandbox/
  package.json          # {"scripts": {"test": "true"}}
  apps/id/go.mod        # module example.com/t490/id
  apps/id/main.go
  apps/id/probe.go      # STAGED
```

A PATH shim (`go` wrapper logging invocations, then `exec`ing the real toolchain) establishes whether the gate's toolchain steps actually execute — the gate's green-path steps are silent, so absence-of-output alone would be a vacuous signal.

**T0 — benign `.go` staged, shipped gate, run from repo top:**

```
$ git add apps/id/probe.go   # comment-only change, vets clean
$ moai gate ; echo EXIT=$?
graph-freshness: ... (advisory)
typecheck: skipped (no scripts.typecheck and no tsconfig.json; ...)
ast-grep scan skipped: ...

quality gate steps (3 configured):
  - typecheck: skipped — ...
  - eslint: skipped — ...
  - npm test: executed in 314ms — npm test -- --passWithNoTests
EXIT=0
```

Shim log: **empty** — `go` was invoked **zero** times. Only the Node toolchain ran.

**T1 — the RED control: a real Go defect staged, same shipped gate:**

```go
// apps/id/probe.go — staged
func probe() string { return gateProbeBug }   // undefined: gateProbeBug
```

`go vet` run directly in `apps/id` catches it (`vet: ./probe.go:5:9: undefined: gateProbeBug`, exit 1) — the defect is real and the Go toolchain present on PATH would fail it. The gate's answer:

```
$ moai gate ; echo EXIT=$?
... quality gate steps (3 configured): ... npm test: executed ...
EXIT=0          # the defect passes silently; shim log still empty
```

**T2 — other languages share the defect.** Same defect class, live-proven for Python (ruff + mypy installed): a repo whose only marker is a nested `packages/api/pyproject.toml` plus a staged `broken.py` carrying `F821 Undefined name` (ruff run directly: exit 1):

```
$ moai gate ; echo EXIT=$?
graph-freshness: ... (advisory)
EXIT=0          # no toolchain detected at all — no steps section, no ruff, no mypy
```

At the source level the exposure is language-agnostic by construction: all 16 toolchains flow through the same single-level marker loop (`go.mod`, `package.json`, `pyproject.toml`/`setup.py`/`requirements.txt`, `Cargo.toml`, `pom.xml`/`build.gradle*`, `*.csproj`/`*.sln`, `Gemfile`, `composer.json`, `Package.swift`, `pubspec.yaml`, `mix.exs`, `build.sbt`, `*.cabal`/`stack.yaml`, `build.zig`). Any of them nested below the project top is invisible.

**Contrast (level is the causal variable):** the same installed binary, same real repository, run from `apps/id/` — where the marker sits at the resolved level — detects Go and runs it (go vet 1.3s + go test 6.3s, exit 0). Nothing is wrong with the Go toolchain steps; only the scan level is wrong.

## Root cause

`internal/hook/quality/gate.go`:

- `detectToolchain()` — markers tested only at `filepath.Join(dir, marker)` for the resolved project dir; returns the first hit; `nil` when none.
- `Run()` — `tc == nil` → `return true, gfNotice` (fail-open by design for genuinely non-software trees, but indistinguishable from "software tree whose markers live one level down").
- `runStep()` — `cmd.Dir = projectDir`, so step arguments (`./...`) resolve at the top, not at the module that owns them.

## Proposed repair (prototype attached, behavior verified)

Not the staged-file reverse traversal #1679 used for the hook — the heavy gate is project-scoped and index-blind, so there are no staged paths to ascend from. The analogous-but-different shape:

1. **`detectToolchains()`** — keep the root-level first match (existing precedence), then `filepath.WalkDir` below the project dir (bounded depth, skipping `sourceScanSkipDirs` dependency/build/cache dirs) collecting **one entry per language** whose marker matches; a dir that matched a marker is not descended into (what lives below a module root belongs to it).
2. **`Run()`** loops over all detected entries — vet/typecheck/lint per toolchain, then the project-level ast-grep step once, then each toolchain's test step.
3. **`executeStep()`/`runStep()`** take the toolchain's own root as `cmd.Dir` (empty falls back to the project dir, preserving existing behavior for root-level projects); `configFiles` existence and `sourceExts` scans scope to that root too.

Why detection alone would make things *worse*: without (3), the newly-detected nested Go runs `go vet ./...` from the project top → `cannot find main module` → every commit red. The three layers land together or the failure mode just flips direction.

Prototype diff (against current `main`, `internal/hook/quality/`): +235/−87 across `gate.go`, plus mechanical signature updates in 5 test files. `go test ./internal/hook/quality/` passes on the patched tree (`ok 5.808s`). The full diff ships in the report bundle; the load-bearing new function:

```go
// detectToolchains: root-level first match preserves existing precedence;
// then a bounded walk below the project dir collects one entry per language.
found = append(found, detectedToolchain{tc: ..., root: path})  // root = dir whose marker matched
```

### Verification matrix (sandbox A/B, same source tree, only the patch differing)

| Cell | Layout | Binary | Expected | Observed |
|---|---|---|---|---|
| T0 | root pkg.json + nested go.mod, benign `.go` staged | pre-patch | reproduce defect | exit 0; Node steps only; `go` invoked 0× ✔ |
| T1 | same, real Go defect staged | pre-patch | defect passes silently | exit 0; `go` invoked 0× ✔ |
| T2 | nested-only pyproject + broken `.py` | pre-patch | no toolchain detected | exit 0; no steps section ✔ |
| T0' | same as T0 | **patched** | Go stage appears, green | exit 0; shim log: `go vet ./...` + `go test ./...` ✔ |
| T1' | same as T1 | **patched** | RED with verbatim diagnostic | exit 1; `vet: ./probe.go:5:9: undefined: gateProbeBug` ✔ |
| T2' | same as T2 | **patched** | nested Python detected, ruff RED | exit 1; `F821 Undefined name 'probe_bug'` ✔ |

Known display delta, not a regression: current `develop`-line builds print a `quality gate steps (N configured):` summary table that `main`-line sources no longer contain; the A/B pair above is main-vs-main±patch so the summary is absent from both sides (per-axis skip notices carry the same information).

## Environment

- moai-adk v3.1.3-rc.5 (build 22df80e90, 2026-08-27); `internal/hook/quality/` byte-identical between that commit and current `main` (48239c7dc) — the mechanism above is what ships
- macOS darwin/arm64, go 1.26.0 (homebrew), node v22.14.0 / npm 11.2.0, ruff + mypy installed, cargo/pytest absent (optional-step skip path)

GoosLab:
# fix(quality): detect toolchain markers below project top and bind module roots into step execution

> **PR draft — NOT SUBMITTED.** Submission is gated on operator approval (구스).
> Branch: `fix/heavy-gate-nested-toolchain` @ `70663c09b` (base `main` @ `48239c7dc`).

**Fixes #1680** · Companion context: #1679 (pre-commit fast-subset — different component, different repair surface; see "Boundary with #1679" below).

## Summary

The heavy gate (`moai gate`, `internal/hook/quality`) tests each language's marker files at exactly one path — the project top. A repository whose Go module (or any of the other 15 toolchains' markers) lives below the top takes the fail-open path: `detectToolchain()` returns `nil`, `Run()` returns `true, gfNotice`, and every commit passes with **zero** toolchain coverage — real defects included. Unlike #1679 (which failed loudly and forced workarounds), this defect passes quietly and skips the inspection the gate exists to perform.

The defect is three-layered, so the repair lands all three layers together:

1. **`detectToolchains()`** — keeps the root-level first match (existing precedence), then runs a bounded `filepath.WalkDir` below the project dir (skips `sourceScanSkipDirs` dependency/build/cache dirs, one entry per language, a matched dir is not descended into) collecting `(toolchain, root)` entries.
2. **`Run()`** — loops over all detected entries: vet/typecheck/lint per toolchain, then the project-level ast-grep step once, then each toolchain's test step.
3. **`executeStep()`/`runStep()`** — take the toolchain's own root as `cmd.Dir` (empty falls back to the project dir, preserving existing behavior for top-level projects); `configFiles` existence and `sourceExts` scans scope to that root too.

Why detection alone would make things *worse*: without (3), the newly-detected nested Go module would run `go vet ./...` from the project top → `cannot find main module` → every commit red. The three layers land together or the failure mode just flips direction from silent-pass to loud-fail.

## Reproduction (see #1680 for the full matrix)

Sandbox mirroring the triggering shape (root `package.json` + nested `apps/id/go.mod`, benign or defective `.go` staged), run with a PATH shim (`go` wrapper logging invocations, then exec'ing the real toolchain — the gate's green-path steps are silent, so absence-of-output alone would be a vacuous signal):

- **Pre-patch (shipped binary, v3.1.3-rc.5):** benign staged `.go` → exit 0, `go` invoked 0×; a real vet defect (`undefined: gateProbeBug`) → exit 0, `go` invoked 0× — the defect passes silently.
- **Level-is the-causal-variable control:** the same binary run from `apps/id/` (where the marker sits at the resolved level) detects Go and runs it green.

## Verification on current `main` (re-measured 2026-08-28, `main` @ `48239c7dc`)

Patches applied to a fresh clone of current `main` (clean apply, 0 conflicts, 6 files, +237/−82). Full A/B re-measurement on this exact tree — same source, only the binary differing:

| Cell | Staging | Binary | Result |
|---|---|---|---|
| T0-M | benign `.go` | unpatched `main` build | exit 0; `go` invoked 0× (shim log absent) ✔ reproduces |
| T1-M | real vet defect | unpatched `main` build | exit 0 — **silent pass reproduced on current main** ✔ |
| T0-P | benign `.go` | patched build | exit 0; shim log: `go vet ./...` + `go test ./...` **actually invoked** ✔ |
| T1-P | real vet defect | patched build | exit 1; verbatim `vet: ./probe.go:5:9: undefined: gateProbeBug` ✔ |

Direct-toolchain controls (proving the staged defects/toolchain are what they claim): positive cell `go vet ./...` + `go test ./...` exit 0; defect cell `go vet ./...` exit 1 with the same verbatim diagnostic.

Regression suite on the patched tree: `go test ./internal/hook/quality/` → `ok 6.662s`; consumers `go test ./internal/cli/` → `ok 241.807s`, `go test ./internal/hook/` → `ok 26.483s`; `go vet ./...` and `go build ./...` green. No dedicated CI guard in this repository touches these files (`manager_lead_depth_test.go` lives in the downstream mo.ai.kr harness, not upstream — checked, 0 hits).

## Boundary with #1679

#1679 owns the **pre-commit fast-subset** (shell template, staged-file reverse traversal). This PR owns the **heavy gate** (`internal/hook/quality`, Go source, project-scoped index-blind walk). Same root theme — "top-level-only placement assumptions" — different component, different repair surface, different verification matrix; hence a separate issue (#1680) and separate PR, cross-referenced both ways.

## Scope notes / known limits (prototype → production)

- One module per language (deepest-unreachable rule: first/shallowest marker wins, matched dirs are not descended into); multiple same-language modules are not enumerated.
- Walk depth is bounded (constant) — repositories with modules deeper than the cap are still invisible (fails open, same as today).
- The walk skips `sourceScanSkipDirs` to keep cost bounded on large trees.
- CI matrix verification (darwin/windows/linux) remains for upstream CI after submission.

🗿 MoAI

## Summary by CodeRabbit

* **New Features**
  * Quality checks now detect and evaluate multiple language toolchains within a project, including nested modules.
  * Static analysis and tests run using each module’s own project directory and configuration.
  * Domain-rule scanning runs consistently across multi-toolchain projects.

* **Bug Fixes**
  * Improved quality-check accuracy for nested modules and their staged files.
  * Corrected working-directory handling so checks reference the appropriate source files and configuration.

### DIFF
--- a/internal/hook/quality/gate.go
+++ b/internal/hook/quality/gate.go
@@ -370,60 +370,33 @@ func (g *QualityGate) Run(ctx context.Context) (bool, string) {
 		return false, gfBlockReason
 	}
 
-	tc := g.detectToolchain()
-	if tc == nil {
+	tcs := g.detectToolchains()
+	if len(tcs) == 0 {
 		// No recognized language — pass, carrying the graph-freshness notice
 		// (the step ran before detection; dropping its notice here is the
 		// silence REQ-GF-005 forbids).
 		return true, gfNotice
 	}
 
-	// Step 1: vet steps
-	var vetReason string
-	for _, step := range tc.vetSteps {
-		ok, out := g.executeStep(ctx, step, g.config.VetTimeout)
-		if !ok {
-			return false, out
-		}
-		vetReason = appendReason(vetReason, out)
-	}
-
 	// passReason carries a passing step's notice out to Run's caller. Dropping
 	// it here is what left an absent ast-grep scanner indistinguishable from a
 	// clean scan: the step reported the skip and this frame threw it away.
 	// The graph-freshness notice (step 0) leads: it is the outermost wrap.
-	passReason := appendReason(gfNotice, vetReason)
+	passReason := gfNotice
 
-	// Step 1.5: typecheck axis.
-	//
-	// It runs after vet and before lint so a type error surfaces before the
-	// slower style pass, and so a project whose linter is absent still gets a
-	// correctness gate. Every outcome is reported: a skip is not a failure,
-	// but it is never silent — that silence is what let a broken build through.
-	if g.config.TypecheckEnabled {
-		step, reason, ok := resolveTypecheckStep(tc.typecheckStep, resolveQualityProjectDir(*g.config, "QualityGate.Run.typecheck"), g.config.TypecheckCommand)
-		switch {
-		case ok:
-			passed, out := g.executeStep(ctx, step, g.config.TypecheckTimeout)
-			if !passed {
-				return false, out
-			}
-			passReason = appendReason(passReason, out)
-		default:
-			passReason = appendReason(passReason, reason)
-		}
-	}
-
-	// Step 2: lint steps
-	for _, step := range tc.lintSteps {
-		ok, out := g.executeStep(ctx, step, g.config.LintTimeout)
+	// A monorepo can carry several language toolchains at once (package.json
+	// at the root plus go.mod under apps/id/, say); each detected entry runs
+	// its steps rooted at the directory whose marker matched it.
+	for i := range tcs {
+		out, ok := g.runToolchainStaticSteps(ctx, &tcs[i])
 		if !ok {
 			return false, out
 		}
 		passReason = appendReason(passReason, out)
 	}
 
-	// Step 2.5: ast-grep domain rules
+	// Step 2.5: ast-grep domain rules (project-level — runs once, not per
+	// toolchain; the rules are language-agnostic domain patterns)
 	// ASTG-UPGRADE-001: switched to RunAstGrepGateV2 which uses the unified Scanner
 	if g.config.AstGrepGate != nil && g.config.AstGrepGate.Enabled {
 		// REQ-HCWA-007: route cwd resolution through resolveQualityProjectDir.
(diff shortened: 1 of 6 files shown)


GoosLab (commented):
## Strong recommendation (would-be change request)

**S2 — add dedicated tests for the new logic.** The diff carries zero `+func Test` additions; all five test-file changes (+18/−18 total) are mechanical signature adaptations to the new `executeStep(ctx, step, dir, timeout)` shape. The new paths have no direct coverage:

- `detectToolchains()` WalkDir nested scan (`internal/hook/quality/gate.go`, new function ~L497+): nested marker discovery below the project top — the defect #1680 reports — plus the per-language-once `detected` map.
- `nestedScanDepthCap = 6` boundary (rel-separator counting → `fs.SkipDir`).
- Skip rules: `sourceScanSkipDirs` pruning, no descent below a matched module root, unreadable-subtree skip (`err != nil → return nil`).
- Multi-toolchain step loop: `runToolchainStaticSteps` / `runToolchainTestStep` rooted at each detected module dir (root `package.json` + nested `go.mod` both running their own steps).

Suggested minimum: one table-driven test on a temp monorepo (root marker + nested module marker) asserting both toolchains are detected with the correct roots and that steps execute at the module root.

## Informational (non-blocking)

1. **New silent-skip surface**: configFiles existence is now scoped to the toolchain root (`anyConfigFileExistsIn(dir, …)`); a nested module whose lint/typecheck config lives only at the project top silently skips that step. If intended ("bind module roots"), consider emitting a skip notice so it stays distinguishable from a clean run.
2. **Doc accuracy**: "the shallowest match wins" is not guaranteed — WalkDir's lexical DFS can let a deeper, lexically-earlier path claim a language before a shallower sibling. Impact ~none (one entry per language still); consider rewording.
3. **Stale body header**: the PR body still reads "PR draft — NOT SUBMITTED" while the PR is open.
4. **Base branch**: targets `main`; the post-2026-08-27 gitflow doctrine routes card work through `develop` (main via release PRs only). Created on transition day — worth a deliberate choice.

CI note: the darwin/amd64 build failure was a module-proxy transient; the failed-jobs rerun of run 33090462613 completed green (`conclusion: success`), so the only remaining red is the non-required graph-freshness base drift (codemaps layer 45>40).

(posted as a comment: GitHub rejects formal REQUEST_CHANGES from the PR author's own account)

🗿 MoAI

